PR #1, opened 2017-07-21
ZGrab2 Framework

ajholland:
DO NOT MERGE

This is my proposed framework for zgrab2. This is by no means complete and is missing parts along with containing some rough edges (commented out code).

The main goals of this are to ensure a sane commandline, ability to run multiple scans at once, and ability to easily add new protocols. Below are examples of this:
`./zgrab2 --help`
`./zgrab2 http --help`
`./zgrab2 http --endpoint="/.well-known"`
`./zgrab2 tls --user-agent="bob"`
`./zgrab2 -o zgrabOutput http -p 70` 
`./zgrab2 mult -c mult.ini`

Defaults are set on structs as well as short and long names. Short names (one character) are only for upper level options. The main idea is that the same struct that is used for the options of a protocol will also contain the fields for results of the protocol. This struct is marshaled and output at the end. This achieves the goal of easily adding new protocols. By specifying short names/long names it becomes an option, otherwise it is a field that will be added by the protocol and output at the end. 

At present this program accepts no input but pretends as if 2 IP's have been entered into a process queue.


zlib/config.go
```diff
@@ -0,0 +1,144 @@
...
+	log "github.com/sirupsen/logrus"
+	"os"
+	"strings"
+)
+
+type Option struct {
+	OutputFileName     string     `short:"o" long:"output-file" default:"-" description:"Output filename, use - for stdout" json:"out"`
+	InputFileName      string     `short:"i" long:"input-file" default:"-" description:"Input filename, use - for stdin" json:"in"`
+	Meta               string     `short:"m" long:"metadata-file" default:"-" description:"Metadata filename, use - for stdout" json:"meta"`
+	Log                string     `short:"l" long:"log-file" default:"-" description:"Log filename, use - for stdout" json:"log"`
```
zakird: This should probably be named consistently with OutputFileName and InputFileName as MetadataFileName. The same for Log.

zlib/config.go
```diff
@@ -0,0 +1,144 @@
...
+	LookupDomain       bool       `short:"d" long:"lookup-domain" description:"Input contains only domain names" json:"lookup"`
```
zakird: This doesn't really make sense as flag. Why can't we autodetect this?

zlib/config.go
```diff
@@ -0,0 +1,144 @@
...
+	Interface          string     `short:"f" long:"interface" description:"Network interface to send on" json:"inter"`
```
zakird: This should absolutely be `-i` not `-f`.  🔌

zlib/config.go
```diff
@@ -0,0 +1,144 @@
...
+	Timeout            int        `short:"t" long:"timeout" description:"Set connection timeout in seconds" json:"time"`
+	GOMAXPROCS         int        `short:"g" long:"gomaxprocs" default:"3" description:"Set GOMAXPROCS" json:"gomax"`
```
zakird: This should default to OS one not to a static number.

zlib/config.go
```diff
@@ -0,0 +1,144 @@
...
+	Senders            int        `short:"s" long:"senders" default:"1000" description:"Number of send coroutines to use" json:"sender"`
+	ConnectionsPerHost int        `short:"h" long:"connections-per-host" defaults:"1" description:"Number of times to connect to each host (results in more output)" json:"connect"`
+	Prometheus         string     `short:"a" long:"prometheus" description:"Address to use for Prometheus server (e.g. localhost:8080). If empty, Prometheus is disabled." json:"-"`
```
zakird: I probably wouldn't give this a short.

zlib/config.go
ajholland: -f should be input file

zlib/config.go
```diff
@@ -0,0 +1,144 @@
...
+	InputFileName      string     `short:"i" long:"input-file" default:"-" description:"Input filename, use - for stdin" json:"in"`
+	Meta               string     `short:"m" long:"metadata-file" default:"-" description:"Metadata filename, use - for stdout" json:"meta"`
+	Log                string     `short:"l" long:"log-file" default:"-" description:"Log filename, use - for stdout" json:"log"`
+	LookupDomain       bool       `short:"d" long:"lookup-domain" description:"Input contains only domain names" json:"lookup"`
+	Interface          string     `short:"f" long:"interface" description:"Network interface to send on" json:"inter"`
+	Timeout            int        `short:"t" long:"timeout" description:"Set connection timeout in seconds" json:"time"`
+	GOMAXPROCS         int        `short:"g" long:"gomaxprocs" default:"3" description:"Set GOMAXPROCS" json:"gomax"`
+	Senders            int        `short:"s" long:"senders" default:"1000" description:"Number of send coroutines to use" json:"sender"`
+	ConnectionsPerHost int        `short:"h" long:"connections-per-host" defaults:"1" description:"Number of times to connect to each host (results in more output)" json:"connect"`
```
zakird: no short

zlib/config.go
```diff
@@ -0,0 +1,144 @@
...
+
+	for scanner.Scan() {
+		if scanner.Text() != "" {
+			bar := strings.NewReader("[" + scanner.Text())
+			parser := flags.NewParser(&Options[NumProtocols], flags.Default)
+			foo := flags.NewIniParser(parser)
+			if err := foo.Parse(bar); err != nil {
+				log.Fatal(err)
+			}
+			switch scanner.Text()[:3] { //todo: remove this shit, fork goflags or recreate
```
zakird: This needs to be dynamic.

zlib/controller.go
```diff
@@ -0,0 +1,53 @@
+package zlib
+
+// Controller implements worker interface
+type Controller struct {
```
zakird: Rename to something specific to metadata file. E.g., monitor

zlib/controller.go
```diff
@@ -0,0 +1,53 @@
...
+	success  uint
```
zakird: These should probably be plural since singular sounds like a boolean.

zlib/processing.go
```diff
@@ -0,0 +1,102 @@
...
+	outputDone.Add(1)
+
+	// Start the output encoder
+	go func() {
+		for result := range outputQueue {
+			if _, err := out.Write(result); err != nil {
+				panic(err.Error())
+			}
+			if _, err := out.Write([]byte("\n")); err != nil {
+				panic(err.Error())
```
zakird: Fatal

zlib/processing.go
```diff
@@ -0,0 +1,102 @@
...
+	var workerDone sync.WaitGroup
+	var outputDone sync.WaitGroup
+	workerDone.Add(int(workers))
+	outputDone.Add(1)
+
+	// Start the output encoder
+	go func() {
+		for result := range outputQueue {
+			if _, err := out.Write(result); err != nil {
+				panic(err.Error())
```
zakird: Fatal

zlib/processing.go
```diff
@@ -0,0 +1,102 @@
...
+			}
+		}
+		outputDone.Done()
+	}()
+	//Start all the workers
+	for i := 0; i < workers; i++ {
+		go func() {
+			for obj := range processQueue {
+				//divide up, run, consolidate
+				result := GrabWorker(obj)
```
zakird: `GrabWorker` makes me think that you were grabbing a worker. You might want to name this something like `MakeZGrabWorker`

zakird: Serialize/Linearize Grab orker.

There should only be --senders numbers of works and they should process things in a single thread.

mult.ini
```diff
@@ -0,0 +1,10 @@
+[http]
```
zakird: allow specify continue on failure.

zlib/config.go
```diff
@@ -0,0 +1,144 @@
...
+	err = enc.Encode(src)
+	if err != nil {
+		return err
+	}
+	dec := gob.NewDecoder(r)
+	return dec.Decode(dst)
+}
+
+// this is a super hacky method of ensuring that all options get defaults set
+func deepCopyAll() {
```
zakird: This will go away with new library.

zlib/processing.go
```diff
@@ -0,0 +1,102 @@
...
+type Handler func(interface{}) interface{}
+
+// not good name, should change
+type protocolResponse struct {
+	protocol string
+	result   interface{}
+}
+
+// GrabWorker divides up input and sends to each handler and then consolidates at end
+func GrabWorker(input interface{}) []byte {
```
zakird: (rename function)

zlib/processing.go
```diff
@@ -0,0 +1,102 @@
...
+	bufChan := make(chan protocolResponse, NumProtocols)
+
+	protocolResult := make(map[string]interface{})
+
+	for i := 0; i < NumProtocols; i++ {
```
zakird: Actions

zlib/ssh.go
```diff
@@ -0,0 +1,35 @@
...
+	Client            string `long:"client" description:"Mimic behavior of a specific SSH client" json:"client"`
+	KexAlgorithms     string `long:"kex-algorithms" description:"Set SSH Key Exchange Algorithms" json:"kex"`
+	HostKeyAlgorithms string `long:"host-key-algorithms" description:"Set SSH Host Key Algorithms" json:"hostkey"`
+	NegativeOne       bool   `long:"negative-one" description:"Set SSH DH kex value to -1 in the selected group" json:"negativeone"`
+
+	scan bool
+}
+
+// Execute validates the options sent to SSHConfig and then passes operation back to main
+func (x *SSHConfig) Execute(args []string) error {
```
zakird: name this better

zlib/tls.go
```diff
@@ -0,0 +1,40 @@
+package zlib
+
```
zakird: There needs to be

(1) a specific flag validation call
(2) a per-module init call
(3) a per-module per-routine init call

cdzombak:
It might be worthwhile to update the `README` on this branch with a quick overview of how configuration works. Writing documentation as we go should help us avoid an outcome like the current state of zgrab1 documentation (https://github.com/zmap/zgrab/issues/132, https://github.com/zmap/zgrab/issues/129, https://github.com/zmap/zgrab/issues/78, https://github.com/zmap/zgrab/issues/31, https://github.com/zmap/zgrab/issues/28, https://github.com/zmap/zgrab/issues/3).

> The main idea is that the same struct that is used for the options of a protocol will also contain the fields for results of the protocol. This struct is marshaled and output at the end. This achieves the goal of easily adding new protocols. By specifying short names/long names it becomes an option, otherwise it is a field that will be added by the protocol and output at the end.

Without having read the code, I don't see what reusing this struct gives us. My mental model, at a high level, is something like: "Options is parsed into a struct, which we treat as immutable once the options have been parsed. That struct is used as input for the grabber, which produces some JSON-serializable structs as output."

Why doesn't that work? How does reusing this one struct "achieve the goal of easily adding new protocols"?

cdzombak:
Looking at the organization of this project — again, before digging into the code — it seems like you have protocol definitions (`tls.go`, `ssh.go` etc) under `zlib` alongside `config.go` and `processing.go`. Especially when considering that we may want to add many, many more protocols, it would make sense to better organize this project so the zgrab core is clearly separated from protocol-specific files.

I know @zakird suggested some package reorganization in his review, too.

ajholland:
@cdzombak 

I'll make sure to keep the readme/docs up to date once we get this stuff ironed out. 

In Zgrab we have two massive structs, Config and GrabData. In order to add a new protocol a user has to define a config struct and add it to Config and then define a separate log/grab struct and add that to GrabData. 

By making it one struct we reduce the number of structs the user has to define and the different places it has to be defined. 

I can see where that would be confusing and an unneeded simplification, I'll change that. 

+1 On the package reorganization

zlib/controller.go
cdzombak: Yep, this definitely needs a clearer name.

main.go
```diff
@@ -0,0 +1,23 @@
...
+func main() {
+	if _, err := parser.Parse(); err != nil {
+		if flagsErr, ok := err.(*flags.Error); ok && flagsErr.Type == flags.ErrHelp {
+			os.Exit(0)
+		} else {
+			log.Fatal("Error: ", err.Error())
+		}
+	}
+
+	c := zlib.MakeNewController()
```
cdzombak: In line with https://github.com/zmap/zgrab2/pull/1#discussion_r1288295481 and the `MakeHandler` function elsewhere, this function should just be called `MakeController` — let's be consistent with naming conventions (in this case, `MakeThing` instead of `MakeNewThing`).

cdzombak:
@ajholland as always, happy to discuss any feedback in person, too. I'll be at BBB tomorrow.

main.go
ajholland: Main package named zgrab or zgrab2?

dadrian: zgrab2 because that's the git repository. You can import it as zgrab if you want. We'll fix this all up once zgrab1 is dead.

ajholland:
If any of you have time I'd appreciate a review again. 

Current status: 
Added input & dynamically parsing cidr/ip/domain
Added zflags & modifications to flags library, currently hosted at github.com/ajholland/zflags
Added ability to add module to zproto without editing a zgrab file
Added lookup

Need to add:
- [ ] refactor names, i.e. zproto? BaseProtocol or BaseModule? protocolResult/protocolResponse, **xssh/ssh?**
- [ ] dns lookup of domain
- [ ] conn abstraction/connect to ip
- [ ] finish implementing ssh, http, tls banner gets

zakird:
Let's schedule time on Tuesday to go over this.

README.md
```diff
@@ -1,4 +1,85 @@
...
+endpoint="/"
+[ssh]
+port=22
+```
+```
+./zgrab2 mult -c mult.ini
+```
+`Application Options` must be the initial section name. Other section names should correspond exactly to the relevant zgrab2 module. The default name for each module is the command name. If the same module is to be used multiple times then `name` must be specified and unique. 
+
+## Contributing
```
zakird: I'd probably change this to "Adding New Protocols"

main/main.go
```diff
@@ -0,0 +1,30 @@
...
+	"os"
+
+	"github.com/ajholland/zflags"
+	log "github.com/sirupsen/logrus"
+	"github.com/zmap/zgrab2/zgrab2"
+	_ "github.com/zmap/zgrab2/zmodules"
+)
+
+func main() {
+	if _, err := zgrab2.ParseFlags(); err != nil {
```
zakird: I'd probably name the package `zgrab` and not `zgrab2`.

zgrab2/processing.go
```diff
@@ -0,0 +1,124 @@
...
+				log.Fatal(err.Error())
+			}
+		}
+		outputDone.Done()
+	}()
+	//Start all the workers
+	for i := 0; i < workers; i++ {
+		go func() {
+			for obj := range processQueue {
+				//divide up, run, consolidate
```
zakird: remove comment

zgrab2/processing.go
```diff
@@ -0,0 +1,124 @@
...
+				result := RunGrabWorker(obj, mon)
+				outputQueue <- result
+			}
+			workerDone.Done()
+		}()
+	}
+
+	// Read the input, send to workers
+	input := bufio.NewReader(config.inputFile)
```
zakird: Probably should be an argument to `Process`.

zgrab2/processing.go
```diff
@@ -0,0 +1,124 @@
...
+	for {
+		obj, err := input.ReadBytes('\n')
```
ajholland: support different inputs: 
ip
domain
cidr block
ip, domain <- not implemented yet

zgrab2/processing.go
ajholland: pass in as reader

zgrab2/processing.go
```diff
@@ -0,0 +1,124 @@
...
+		if err == io.EOF {
+			break
+		} else if err != nil {
+			log.Error(err)
+		}
+		st := string(obj)
```
zakird: Check that this doesn't break something unicode.

zgrab2/processing.go
```diff
@@ -0,0 +1,124 @@
...
+		ipnet, domain, err := ParseInput(st[:len(st)-1]) //remove newline
+		if err != nil {
+			log.Error(err)
+		} else {
+			if domain == "" {
+				for _, ip := range ipnet {
```
zakird: need to handle ip,domain case.

zgrab2/processing.go
```diff
@@ -0,0 +1,124 @@
...
+		if err == io.EOF {
+			break
+		} else if err != nil {
+			log.Error(err)
+		}
+		st := string(obj)
+		ipnet, domain, err := ParseInput(st[:len(st)-1]) //remove newline
+		if err != nil {
+			log.Error(err)
+		} else {
```
zakird: probably shouldn't be an else

zgrab2/processing.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	// Read the input, send to workers
+	input := bufio.NewReader(config.inputFile)
+	for {
+		obj, err := input.ReadBytes('\n')
+		if err == io.EOF {
+			break
+		} else if err != nil {
+			log.Error(err)
```
zakird: put a continue

zgrab2/utility.go
```diff
@@ -0,0 +1,63 @@
...
+		var ips []net.IP
+		for ip := ip.Mask(ipnet.Mask); ipnet.Contains(ip); inc(ip) {
+			ips = append(ips, dupIP(ip))
+		}
+
+		return ips, "", nil
+	}
+}
+
+func inc(ip net.IP) {
```
zakird: name this something useful sounding :p

zgrab2/utility.go
```diff
@@ -0,0 +1,63 @@
...
+	for j := len(ip) - 1; j >= 0; j-- {
+		ip[j]++
+		if ip[j] > 0 {
+			break
+		}
+	}
+}
+
+func dupIP(ip net.IP) net.IP {
```
zakird: give this a better name :|

zmodules/http.go
```diff
@@ -0,0 +1,73 @@
...
+	ProxyRequest  *HTTPRequest  `json:"connect_request,omitempty"`
+	ProxyResponse *HTTPResponse `json:"connect_response,omitempty"`
+	//Response              *http.Response   `json:"response,omitempty"`
+	//RedirectResponseChain []*http.Response `json:"redirect_response_chain,omitempty"`
+}
+
+// Per module initialization call
+func init() {
+	var httpConfig HTTPConfig
+	cmd, err := zgrab2.AddCommand("http", "HTTP Banner Grab", "Grab a banner over HTTP", &httpConfig)
```
zakird: I might name this something like register scanner or the like

zmodules/http.go
```diff
@@ -0,0 +1,73 @@
...
+// Per module per goroutine initialization call
+func (x HTTPConfig) Initialize() {
+
+}
+
+// Validates the options sent to HTTPConfig, registers the config module, and then passes operation back to main
+func (x *HTTPConfig) Validate(args []string) error {
+	zgrab2.ValidateHighLevel()
+
+	zgrab2.RegisterLookup(x.Name, *x)
```
zakird: Figure out how to pull these two commands out o the module and into the framework.

zmodules/http.go
```diff
@@ -0,0 +1,73 @@
...
+	return nil
+}
+
+func (x HTTPConfig) GetBanner() (interface{}, error) {
```
zakird: I might name this something more generic like `Scan()` since not all protocols have banners.

zmodules/http.go
```diff
@@ -0,0 +1,73 @@
+package zmodules
+
+import (
+	log "github.com/sirupsen/logrus"
+	"github.com/zmap/zgrab2/zgrab2"
+)
+
+type HTTPConfig struct {
```
zakird: I might name this "HTTPScanner" or "HTTPModule".

zmodules/ssh.go
```diff
@@ -0,0 +1,40 @@
+package zmodules
+
+import (
+	log "github.com/sirupsen/logrus"
+	"github.com/zmap/zgrab2/zgrab2"
+)
+
+type SSHConfig struct {
```
zakird: Get opinion from @dadrian on segmenting flags out into a sub-struct.

zmodules/ssh.go
```diff
@@ -0,0 +1,40 @@
...
+	var sshConfig SSHConfig
+	cmd, err := zgrab2.AddCommand("ssh", "SSH Banner Grab", "Grab a banner over SSH", &sshConfig)
+	if err != nil {
+		log.Fatal(err)
+	}
+	sshConfig.SetDefaultPortAndName(cmd, uint(22), "ssh")
+}
+
+// per module per routine initialization call
+func (x SSHConfig) Initialize() {
```
zakird: might just name this `PerRoutineInitialize`.

zgrab2/processing.go
ajholland: `process` is in the same package as `config`. Should they both just be removed from the arguments list and used in this form? I don't see the point of calling them from main (i.e. zgrab2.Process(zgrab2.config.inputFile))

zakird:
I could see either way. @dadrian thoughts?

zgrab2/handler.go
```diff
@@ -0,0 +1,17 @@
...
+	res, e := module.Scan()
+	var err *error //nil pointers are null in golang, which is not nil and not empty
+	if e == nil {
+		mon.statuses <- status_success
+		err = nil
+	} else {
+		mon.statuses <- status_failure
+		err = &e
+	}
+	resp := moduleResponse{Result: res, Error: err}
```
dadrian: I'm confused. Interfaces are already effectively pointers, why not just pass the error directly? There's no situation where resp.Error is nil and e isn't also nil.

zgrab2/interface.go
```diff
@@ -0,0 +1,50 @@
...
+func (b BaseModule) GetName() string {
+	return b.Name
+}
+
+func (b *BaseModule) SetDefaultPortAndName(cmd *flags.Command, port uint, name string) {
+	cmd.FindOptionByLongName("port").Default = []string{strconv.FormatUint(uint64(port), 10)}
+	cmd.FindOptionByLongName("name").Default = []string{name}
+}
+
+var lookups map[string]Module
```
dadrian: This should be a map of string to pointer to module, as discussed offline.

zgrab2/interface.go
```diff
@@ -0,0 +1,50 @@
...
+		lookups = make(map[string]Module, 10)
+	}
+	//add to list and map
+	if lookups[name] != nil {
+		log.Fatal("name already used")
+	}
+	lookups[name] = m
+}
+
+func NumActions() uint {
```
dadrian: This may as well be an int since `len` returns an int.

zgrab2/multiple.go
```diff
@@ -0,0 +1,31 @@
+package zgrab2
+
+import (
+	"errors"
+	"os"
+
+	"github.com/ajholland/zflags"
+)
+
+type MultipleConfig struct {
```
dadrian: Not convinced this struct is named well. A config file doesn't necessarily define multiple scans, does it? Why isn't this part of the other the regular config struct? Can the config file override the global config things like `GOMAXPROCS` or input/output file?

zgrab2/multiple.go
```diff
@@ -0,0 +1,31 @@
...
+	ConfigFileName  string `short:"c" long:"config-file" default:"-" description:"Config filename, use - for stdin"`
+	ContinueOnError bool   `long:"continue-on-error" description:"If proceeding protocols error, do not run following protocols (default: true)"`
+}
+
+// Validates the options sent to MultipleConfig, and parses the configFile
+func (x *MultipleConfig) Validate(args []string) error {
```
dadrian: This config stuff is getting really hard to keep track of. What's the rationale against having a single config struct, a single `ConfigFileName` string, and then just reading the file at the start if passed?

zgrab2/multiple.go
```diff
@@ -0,0 +1,31 @@
...
+	if x.ConfigFileName == config.InputFileName {
```
dadrian: This seems like an excessive error check. Won't reading input just fail in this case?

zgrab2/processing.go
```diff
@@ -0,0 +1,126 @@
...
+	"encoding/json"
+	"io"
+	"net"
+	"sync"
+
+	log "github.com/sirupsen/logrus"
+)
+
+type Grab struct {
+	IP     *string                   `json:"ip,omitempty"`
```
dadrian: Why is this a pointer? It's exceedingly easy to have concurrency bugs with pointers.

dadrian: (I'm assuming you send this through a channel at some point)

ajholland: omitempty for json marshaling only works if the type is set to nil. A string can't be set to nil and the empty string isn't empty according to omitempty. 
It does get sent through a channel

zgrab2/multiple.go
ajholland: I thought we predetermined that a config file **does** necessarily define multiple scans?

And yes, config file override's the global config things. The use cases should be single scan: all options entered through cli. Multiple scan: all options entered through config file, but also supports global config things being entered from cli, however if you specify both (I don't see why) config file takes precedence.

zgrab2/multiple.go
ajholland: The rational would be that someone could specify `./zgrab2 --config-file="mult.ini" http --port=8080`, which would make no sense. Either do a multiple scan, or do a single http. 

I think the naming here is probably the main issue, maybe instead of MultipleConfig it should be MultipleModules?

zgrab2/handler.go
ajholland: Similar case to strings where omitempty wouldn't actually omit an empty err since it would be "null" instead of nil.

zgrab2/config.go
```diff
@@ -0,0 +1,82 @@
...
+	default:
+		var err error
+		if config.metaFile, err = os.Create(config.MetaFileName); err != nil {
+			log.Fatal(err)
+		}
+	}
+
+	// Validate Go Runtime config
+	if config.GOMAXPROCS < 0 {
+		log.Fatal("invalid GOMAXPROCS (must be at least 1, given %d)", config.GOMAXPROCS)
```
ajholland: https://github.com/zmap/zgrab/blame/4c4f01ee32a5b6340c1d3b8f5300502549672e5d/main.go#L140

Past David says it does :p

zgrab2/config.go
```diff
@@ -0,0 +1,82 @@
...
+	//Validate files
+	switch config.LogFileName {
+	case "-":
+		config.logFile = os.Stderr
+	default:
+		var err error
+		if config.logFile, err = os.Create(config.LogFileName); err != nil {
+			log.Fatal(err)
+		}
+		log.SetOutput(config.logFile)
```
ajholland: yup, default is stderr

zgrab2/config.go
```diff
@@ -0,0 +1,82 @@
...
+	Multiple           MultipleConfig `command:"multiple" description:"Multiple module actions"`
+
+	inputFile  *os.File
+	outputFile *os.File
+	metaFile   *os.File
+	logFile    *os.File
+}
+
+func init() {
+	config.Multiple.ContinueOnError = true //set default for multiple value
```
ajholland: When running Multiple modules someone may only want to run a module if the preceding modules for that ip/domain succeeded.

main/main.go
```diff
@@ -0,0 +1,30 @@
+package main
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/ajholland/zflags"
+	log "github.com/sirupsen/logrus"
+	"github.com/zmap/zgrab2/zgrab2"
```
ajholland: 🤷‍♂️  I did it for directory cleanliness but I can bump it all up

README.md
```diff
@@ -1,4 +1,85 @@
...
+    var configStruct ConfigStruct
+    cmd, err := zgrab2.AddCommand("module", "short description", "long description of module", &configStruct)
+    if err != nil {
+        log.Fatal(err)
+    }
+    configStruct.SetDefaultPortAndName(cmd, uint(00), "module")
+}
+```
+
+The `Validate()` function should first make a call to `zgrab2.ValidateHighLevel()` and lastly call `zgrab2.RegisterLookup()`.
```
ajholland: I agree. I got rid of requiring ValidateHighLevel(), the issue is still the RegisterLookup function and corresponding call. Currently, the flag library parser goes through the .ini file, when it reaches the end of a section it calls Validate() on that section. I register the lookup at the end of that Validate so that the config doesn't get overwritten by a following config. The reason the config would be overwritten is because the parser requires pointers to structs. (Line 74 above, has to do with the way they use reflect package, goroutines panic otherwise)
I'm still stuck on two related issues here. A) how to remove the RegisterLookup call, and B) how to change the interfaces to work on pointer receivers.
Suggestions? We discussed earlier about changing the flag library but at this point that would involve refactoring a large part of the library or using the reflect package to deep copy the data from one pointer to a new pointer.

interface.go
```diff
@@ -0,0 +1,51 @@
...
+func (b *BaseModule) SetDefaultPortAndName(cmd *flags.Command, port uint, name string) {
+	cmd.FindOptionByLongName("port").Default = []string{strconv.FormatUint(uint64(port), 10)}
+	cmd.FindOptionByLongName("name").Default = []string{name}
+}
+
+var lookups map[string]*Module
+
+func RegisterLookup(name string, m Module) {
+	if lookups == nil {
+		lookups = make(map[string]*Module, 10)
```
ajholland: Do this in init() and remove hard coded 10

interface.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"log"
+	"net"
+	"strconv"
+
+	"github.com/ajholland/zflags"
+)
+
+type Module interface {
+	Scan(ip net.IP) (interface{}, error)
```
ajholland: In depth comments

processing.go
```diff
@@ -0,0 +1,127 @@
...
+	Domain string                    `json:"domain,omitempty"`
+	Data   map[string]moduleResponse `json:"data,omitempty"`
+}
+
+type grabTarget struct {
+	IP     net.IP
+	Domain string
+}
+
+type moduleResponse struct {
```
ajholland: public

processing.go
```diff
@@ -0,0 +1,127 @@
...
+	Result         interface{} `json:"result,omitempty"`
+	Time           string      `json:"time,omitempty"`
+	Error          *error      `json:"error,omitempty"`
+	ErrorComponent string      `json:"error_component,omitempty"`
+}
+
+// GrabWorker calls handler for each action
+func RunGrabWorker(input grabTarget, m *Monitor) []byte {
```
ajholland: private

ajholland: pointer grabTarget

ajholland: rename grabTarget to target and this function should be grabTarget

processing.go
```diff
@@ -0,0 +1,127 @@
...
+	"encoding/json"
+	"io"
+	"net"
+	"sync"
+
+	log "github.com/sirupsen/logrus"
+)
+
+type Grab struct {
+	IP     *string                   `json:"ip,omitempty"`
```
ajholland: shouldn't be pointer

processing.go
```diff
@@ -0,0 +1,127 @@
...
+	Time           string      `json:"time,omitempty"`
+	Error          *error      `json:"error,omitempty"`
+	ErrorComponent string      `json:"error_component,omitempty"`
+}
+
+// GrabWorker calls handler for each action
+func RunGrabWorker(input grabTarget, m *Monitor) []byte {
+	moduleResult := make(map[string]moduleResponse)
+
+	for _, action := range lookups {
```
ajholland: keep second list of names because map order undefined

processing.go
```diff
@@ -0,0 +1,127 @@
...
+		name, res := runHandler(*action, m, input.IP)
+		moduleResult[name] = res
+		if res.Error != nil && !config.Multiple.ContinueOnError {
+			break
+		}
+	}
+
+	var ipstr *string
```
ajholland: delet

processing.go
```diff
@@ -0,0 +1,127 @@
...
+		ipstr = nil
+	} else {
+		s := input.IP.String()
+		ipstr = &s
+	}
+
+	a := Grab{IP: ipstr, Domain: input.Domain, Data: moduleResult}
+	result, err := json.Marshal(a)
+	if err != nil {
+		log.Fatal(err)
```
ajholland: fatalf

processing.go
```diff
@@ -0,0 +1,128 @@
...
+
+	// Start the output encoder
+	go func() {
+		out := bufio.NewWriter(config.outputFile)
+		defer out.Flush()
+		for result := range outputQueue {
+			if _, err := out.Write(result); err != nil {
+				log.Fatal(err)
+			}
+			if _, err := out.Write([]byte("\n")); err != nil {
```
ajholland: WriteByte

processing.go
```diff
@@ -0,0 +1,128 @@
...
+				log.Fatal(err)
+			}
+		}
+		outputDone.Done()
```
ajholland: defer

processing.go
```diff
@@ -0,0 +1,128 @@
...
+		}
+		st := string(obj)
+		ipnet, domain, err := ParseInput(st[:len(st)-1]) //remove newline
+		if err != nil {
+			log.Error(err)
+			continue
+		}
+
+		if ipnet != nil {
+			for _, ip := range ipnet {
```
ajholland: fix this to not waste a ton of memory, actual net.ipnet

handler.go
```diff
@@ -0,0 +1,23 @@
+package zgrab2
+
+import (
+	"net"
+	"time"
+)
+
+// runHandler will call perRoutineInitialize, Scan, and respond with a protocol response, data unmarshalled, to the worker
+func runHandler(module Module, mon *Monitor, ip net.IP) (string, ModuleResponse) {
```
ajholland: rethink function name

ajholland: RunModule, public

ajholland: rename interface.go to module.go move this to there

interface.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"log"
+	"net"
+	"strconv"
+
+	"github.com/ajholland/zflags"
+)
+
+type Module interface {
```
ajholland: ScanModule

utility.go
```diff
@@ -0,0 +1,86 @@
...
+	i := strings.IndexByte(s, ',')
+	j := strings.IndexByte(s, '/')
+
+	switch {
+	case i == -1 && j == -1:
+		//just ip or domain
+		if ip := net.ParseIP(s); ip != nil {
+			return []net.IP{ip}, "", nil
+		} else {
+			ips, err := net.LookupIP(s)
```
ajholland: @dadrian Currently I parse a cidr block and put it in a list. Bad idea. Trying to remove that and put it as a IPNet instead. I'm blocking on this line here, where I lookup the ip's associated with a domain. Should I just return 4 types, ([]net.IP, string, net.IPNet, error) or only perform a scan on the first IP associated with a domain? Presently zgrab2 will lookup a domain, find all IP's associated with that domain, and then scan all of them.

dadrian: Only scan the first one for now.

PR #9, opened 2017-10-23
Implement SSH

ajholland:
This PR implements the SSH protocol into ZGrab2. It takes the XSSH library from original zgrab and applies a couple fixes to make it work with ZGrab2. 

Changes:
- XSSH to SSH
- Non-global Configuration

The main file of interest here is probably modules/ssh.go

I diffed the output on a couple of hosts and it looks mostly the same % the different output of zgrab2 due to multiple scans.


modules/ssh.go
```diff
@@ -22,10 +34,14 @@ type SSHScanner struct {
...
 	var sshModule SSHModule
-	_, err := zgrab2.AddCommand("ssh", "SSH Banner Grab", "Grab a banner over SSH", 22, &sshModule)
+	cmd, err := zgrab2.AddCommand("ssh", "SSH Banner Grab", "Grab a banner over SSH", 22, &sshModule)
 	if err != nil {
 		log.Fatal(err)
 	}
+	s := ssh.MakeSSHConfig() //dummy variable to get default for host key, kex algorithm, ciphers
+	cmd.FindOptionByLongName("host-key-algorithms").Default = []string{strings.Join(s.HostKeyAlgorithms, ",")}
+	cmd.FindOptionByLongName("kex-algorithms").Default = []string{strings.Join(s.KeyExchanges, ",")}
+	cmd.FindOptionByLongName("ciphers").Default = []string{strings.Join(s.Ciphers, ",")}
```
dadrian: Why are these set here instead of in the struct tag? Is it just because they're long?

modules/ssh.go
```diff
@@ -57,6 +73,44 @@ func (s *SSHScanner) InitPerSender(senderID int) error {
 func (s *SSHScanner) GetName() string {
 	return s.config.Name
 }
-func (s *SSHScanner) Scan(t zgrab2.ScanTarget, port uint) (interface{}, error) {
-	return nil, nil
+
+func (s *SSHScanner) makeSSHGrabber(hlog *ssh.HandshakeLog) func(string) error {
```
dadrian: This probably doesn't need to return a closure and can just be inlined as part of `Scan`.

ajholland: 👍 Will push change in a few

modules/ssh.go
ajholland: Yeah, setting them in the struct tag would just be a long string that I didn't think should be hardcoded. It still shows up in the help docs though. 

I think the answer is to put these long outputs in the Usage part of the help docs, not shown as defaults inline with the flag. And also to set these defaults in the ssh config if alternates are not specified. Does that sound right?

ajholland:
Module structure felt fine, works well and seems like it should allow for flexibility for other modules. The pain point was modifying xssh to remove the global config struct and the flags setup that it already had. But that's just something that has to be done eventually. ¯\_(ツ)_/¯

modules/ssh.go
dadrian: Can you paste what the output of `--help` looks like?

ajholland: zgrab2

PR #23, opened 2017-11-27
add MySQL module

justinbastress:
# What it does:
* Add mysql zgrab2 module
* Generic TLS grab
  * Add tls.go, abstracting TLS configuration / scanning / output
  * Let ZGrab drive the TLS handshake
     1. The module's Scan() method uses the mysql driver
     2. If the server supports TLS, use the zgrab2 TLS APIs to do the scan based on the input flags 
* Add zschema schemas
* Add travis integration
  * Pull down mysql docker containers to test against
  * Run zschema validation on output

# How to test
When changes are published, Travis will automatically run the integration tests, but it's straightforward to run them manually:
1. Pull down any requirements (see `before_install` and `before_scripts` in .travis.yml)
  * e.g. https://github.com/jmespath/jp
2. Run `start_mysql.sh`
3. Run `integration_tests.sh`
4. Run `cleanup_mysql.sh`

Or, to directly run zgrab, set up a mysql container (see e.g. `launch_mysql_container.sh`), and run e.g.
* `echo 127.0.0.1 | zgrab2 mysql --heartbleed --keep-client-logs --heartbeat-enabled`

# Notes / Caveats
* The debug-only tags are present in the mysql structs, but not in zcrypto or ssh
  * There is no custom JSON Marshaller to interpret these, either
* Only the mysql module has integration tests
* The scan statuses are currently very rudimentary (just success/fail)
* The TLS command-line flags are almost certainly not in their final form (e.g. array values take a comma-separated list, which is processed via a naive split)
* Future TODO: Include input configuration
  * It seems like this would be a good candidate for a zgrab-level result...i.e. drop the `ScanFlags` next to the `result`?

PR #24, opened 2017-11-27
Fix Init per Sender

ajholland:
Actually call Init Per Sender for every registered scanner for each goroutine.

### DIFF
--- a/processing.go
+++ b/processing.go
@@ -91,6 +91,10 @@ func Process(mon *Monitor) {
 	//Start all the workers
 	for i := 0; i < workers; i++ {
 		go func() {
+			for _, scannerName := range orderedScanners {
+				scanner := *scanners[scannerName]
+				scanner.InitPerSender(i)
+			}
 			for obj := range processQueue {
 				for run := uint(0); run < uint(config.ConnectionsPerHost); run++ {
 					result := grabTarget(obj, mon)
--- a/processing.go
+++ b/processing.go
@@ -90,7 +90,7 @@ func Process(mon *Monitor) {
 	}()
 	//Start all the workers
 	for i := 0; i < workers; i++ {
-		go func() {
+		go func(i int) {
 			for _, scannerName := range orderedScanners {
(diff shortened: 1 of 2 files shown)


processing.go
```diff
@@ -91,6 +91,10 @@ func Process(mon *Monitor) {
 	//Start all the workers
 	for i := 0; i < workers; i++ {
 		go func() {
+			for _, scannerName := range orderedScanners {
+				scanner := *scanners[scannerName]
```
dadrian: I don't think you want to dereference here. What's the type of `orderedScanners`?

ajholland: `orderedScanners` is just strings corresponding to scanner name. `scanners` is of type `map[string]*Scanner` where Scanner is an interface, and we can't operate on a pointer to an interface.

justinbastress: For my own education, I tried this without dereferencing it, and it fails to compile with "scanner.InitPerSender undefined (type *Scanner is pointer to interface, not interface)", which seems to make sense in this context (if it was a Scanner.(*MySQLScanner) then maybe not...

processing.go
```diff
@@ -91,6 +91,10 @@ func Process(mon *Monitor) {
...
+				scanner.InitPerSender(i)
```
justinbastress: The intent is to call this once for each i in [0, workers), right? If so, the outer for loop may be a problem --
 you would need to increment the shared counter inside the goroutine, right?

I confirmed by running this with a log method in InitPerSender, that it does indeed get called 1000 times, but, say, ten times for senderID=1000, but not at all for senderID=998.

dadrian: Yeah, you'll need to pass `i` as an argument to the anonymous function:

```
for int i: =0; ....
    go func(i int) {
        for .....
    }(i) 
```

dadrian: Good catch @justinbastress!

ajholland: Pushed a change, thanks @justinbastress!

dadrian:
This is covered by #23

PR #25, opened 2017-12-07
TLS scan prototype + Travis integration

justinbastress:
# What it does:
* Generic TLS grab
  * Add tls.go, abstracting TLS configuration / scanning / output
  * Let ZGrab drive the TLS handshake
     1. The module's Scan() method uses the mysql driver
     2. If the server supports TLS, use the zgrab2 TLS APIs to do the scan based on the input flags 
* Add zschema schemas
* Add travis integration
  * Pull down mysql docker containers to test against
  * Run zschema validation on output

# How to test
When changes are published, Travis will automatically run the integration tests, but it's straightforward to run them manually:
1. Pull down any requirements (see `before_install` and `before_scripts` in .travis.yml)
  * e.g. https://github.com/jmespath/jp
2. Run `start_mysql.sh`
3. Run `integration_tests.sh`
4. Run `cleanup_mysql.sh`

Or, to directly run zgrab, set up a mysql container (see e.g. `launch_mysql_container.sh`), and run e.g.
* `echo 127.0.0.1 | zgrab2 mysql --heartbleed --keep-client-logs --heartbeat-enabled`

# Notes / Caveats
* The debug-only tags are present in the mysql structs, but not in zcrypto or ssh
  * There is no custom JSON Marshaller to interpret these, either
* Only the mysql module has integration tests
* The scan statuses are currently very rudimentary (just success/fail)
* The TLS command-line flags are almost certainly not in their final form (e.g. array values take a comma-separated list, which is processed via a naive split)


justinbastress:
- [ ] Rename tls_handshake to tls
- [ ] Move heartbleed flags into their own object (similar to existing zgrab) that is omitted if heartbleed is not checked
- [ ] Include input configuration in output

tls.go
```diff
@@ -0,0 +1,268 @@
...
+	HeartbeatEnabled    bool   `long:"heartbeat-enabled" description:"If set, include the heartbeat extension"`
+	DSAEnabled          bool   `long:"dsa-enabled" description:"Accept server DSA keys"`
+	// TODO: format?
+	ClientRandom string `long:"client-random" description:"Set an explicit Client Random (base64 encoded)"`
+	// TODO: format?
+	ClientHello string `long:"client-hello" description:"Set an explicit ClientHello (base64 encoded)"`
+}
+
+func getCSV(arg string) []string {
+	// TODO: Find standard way to pass array-valued options
```
justinbastress: This is of course even more embarassing after the discussion Wednesday re: reinventing the argparse wheel; unfortunately, I don't see any standard way to get array-like inputs from the command-line.  This probably isn't the best format (and ditto for the others), but this serves as a placeholder for now.

I will go over zgrab1 again to see if I can duplicate what it was doing for these (if indeed they were configurable).

Maybe the best solution would just be to have it be configurable only in the ini?

PR #30, opened 2017-12-26
Implements postgres zgrab2 module

justinbastress:
## What it does ##

* Provides a zgrab2 module for postgres
* Integration tests for postgres zgrab module
* Various fixes / refactoring

## How to test it ##

Travis should pick it up, but you can set up a basic postgres docker image (no SSL) by running
 * `docker run -itd -p 5432:5432 --rm --name zgrab_pg -e POSTGRES_PASSWORD postgres:10.1`

Then, you can run the scanner against localhost:
* `echo "127.0.0.1" | cmd/zgrab2/zgrab2 postgres --timeout 10`

You should see some server-generated strings in the output, e.g. `"FATAL:  unsupported frontend protocol 0.0: server supports 2.0 to 3.0"`

Or, as always, you can test the integration tests by running `pushd integration_tests/postgres && ./setup.sh && ./test.sh && ./cleanup.sh && popd`, or to run them alongside the rest of the tests, `pushd integration_tests &&./setup.sh && ./test.sh && ./cleanup.sh && popd`

## Notes ##

I'll try to call anything important out in comments to the code, but the main changes are:

 * **integration_tests/**:
   * *Cleanup*: Standardize container/image names (e.g. testmysql, sshtest -> zgrab_mysql, zgrab_ssh)
   * *Cleanup*: Consistently use /usr/bin/env bash rather than /bin/bash
   * *Bugfix*: Ensure set -e gets set everywhere it is needed  (this was leading to false test-passes)
   * *Bugfix*: `test.sh`: Catch / log all validation errors
   * *Bugfix*: `test.sh`: Catch schematically-valid but non-successful scans (e.g. status = connection-refused)
   * *New implementation*: `new.sh`: add schema skeleton

 * **postgres/**:
   * *New implementation*: Initial version of scanner
   * *New implementation*: integration tests

* **schemas/**:
  * *New implementation*: `zcrypto.py`: fill out `client_hello`

* **misc**
   * *Refactor*: Move ScanStatus & related to status.go (previously scattered around processing.go, module.go, utility.go)
   * *Refactor*: Add `ScanTarget.Open(*BaseFlags)` method (so every module doesn't need to re-write the same `Dial("${ScanTarget.IP}:${Config.BaseFlags.Port}")` and `SetReadDeadline` / `SetWriteDeadline` code (and, so scanners no longer need to be aware of the host/port/timeout)
   * *Bugfix*: `tls.go`: Better implementation of TLSConnection.GetLog() -- keep the pointer around and populate it as we go


processing.go
```diff
@@ -24,6 +26,11 @@ type ScanTarget struct {
 	Domain string
 }
 
+// ScanTarget.Open connects to the ScanTarget using net.DialTimeout(), getting the port and timeout from flags.
+func (t *ScanTarget) Open(flags *BaseFlags) (net.Conn, error) {
+	return net.DialTimeout("tcp", fmt.Sprintf("%s:%d", t.IP.String(), flags.Port), time.Second*time.Duration(flags.Timeout))
```
justinbastress: I'm wondering if it would be feasible to use this to automatically set the IO timeouts as well -- i.e. return a new net.Conn implementation (wrapping the `net.DialTimeout()` return value) that calls `SetReadDeadline()` etc, since currently that needs to be re-implemented everywhere.

The main risk I see is, some scanners may need more flexibility (wait different times for different ops) -- but in that case, there's nothing preventing anyone from calling net.DialTimeout directly.

integration_tests/test.sh
```diff
@@ -28,14 +30,43 @@ for mod in $(ls); do
...
+            err="schema failure@$protocol/$outfile"
+            if [[ $status -eq 0 ]]; then
+                failures="$err"
+            else
+                failures="$failures, $err"
+            fi
+            status=1
+        else
+            echo "validation of $target succeeded."
+            scan_status=$($ZGRAB_ROOT/jp -u data.${protocol}.status < $target)
```
justinbastress: Previously, failing mysql scans were caught because it was checking the server version in its own tests (and that was null in the case of a failed scan) -- but the SSH/postgres scans simply checked that the zgrab executable exited with status 0, and that the output was schematically valid -- and since e.g. `"status": "connection-refused"` is indeed valid according to the schema, they virtually never failed.

lib/mysql/mysql.go
```diff
@@ -98,12 +95,6 @@ const (
 )
 
 type Config struct {
-	// @TODO: Does it make sense to make Host/Port connection fields, so that Config can be shared across connections?
```
justinbastress: This is all now handled by tls.go and  ScanTarget.Open().

module.go
```diff
@@ -16,22 +16,6 @@ type Scanner interface {
 	Scan(t ScanTarget) (ScanStatus, interface{}, error)
 }
 
-// ScanStatus is the enum value that states how the scan ended.
-type ScanStatus string
```
justinbastress: Moved to status.go

processing.go
```diff
@@ -24,6 +26,50 @@ type ScanTarget struct {
 	Domain string
 }
 
+// scanTargetConnection wraps an existing net.Conn connection, overriding the Read/Write methods to use the configured timeouts
+type scanTargetConnection struct {
+	net.Conn
+	Timeout time.Duration
+}
+
+func (c *scanTargetConnection) Read(b []byte) (n int, err error) {
```
justinbastress: My main concern here is that some scanners may need more flexibility -- say, different timeouts for different reads.

But, my thinking is that in those cases, the scanner is going to need to know about the timeout anyway, so the it can go ahead and open its own connection with net.Dial directly.

test_mysql_all.sh
```diff
@@ -1,10 +0,0 @@
-#!/bin/bash -e
```
justinbastress: Superseded by integration_tests/test.sh

tls.go
```diff
@@ -232,31 +233,29 @@ type TLSLog struct {
...
-	var heartbleed *tls.Heartbleed
-	if z.flags.Heartbleed {
-		heartbleed = z.Conn.GetHeartbleedLog()
-	} else {
-		heartbleed = nil
-	}
-	return &TLSLog{
-		HandshakeLog:  handshake,
-		HeartbleedLog: heartbleed,
+	if z.log == nil {
```
justinbastress: Previously, callers had to be careful to call GetLog() as late as possible, which was awkward (and unnecessary).

schemas/postgres.py
```diff
@@ -0,0 +1,75 @@
...
+# modules/postgres/scanner.go: PostgresResults
+postgres_scan_response = SubRecord({
+    "result": SubRecord({
+        "tls": zgrab2.tls_log,
+        "supported_versions": String(),
+        "protocol_error": postgres_error,
+        "startup_error": postgres_error,
+        "is_ssl": Boolean(required = True),
+        "authentication_mode": postgres_auth_mode,
+        "server_parameters": String(), # TODO FIXME: This is currendly an unconstrained map[string]string
```
justinbastress: This doesn't cause any problems yet because these are only present when the authentication succeeds, which shouldn't ever happen.

The easiest approach would be to just pick out a subset of parameters that we want to track (say, server_version and TimeZone -- or the union of all values returned by our test containers) and throw away the rest.

docker-runner/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+WORKDIR /go/src/github.com/zmap/zgrab2
+
+RUN go get ./...
+RUN go get -t ./...
+
+# This should already be executable, but just in case...
+RUN chmod a+x ./docker-runner/entrypoint.sh
+
+# Build on the container
+RUN cd cmd/zgrab2 && go build && cd ../..
```
dadrian: I think you can do `go build cmd/zgrab2`, but not sure

integration_tests/postgres/cleanup.sh
```diff
@@ -0,0 +1,20 @@
...
+function clean() {
+  CONTAINER_NAME=$1
+  docker stop $CONTAINER_NAME
+}
+
+versions="9.3 9.4 9.5 9.6 10.1"
+types="ssl nossl"
+for version in $versions; do
+  for type in $types; do
+    clean "zgrab_postgres_${version}-${type}"
```
dadrian: Do we need to `docker rm` these as well?

docker-runner/Dockerfile
andrewsardone: **Question** ([_pull request commenting with a purpose_](https://gist.github.com/andrewsardone/3fb9746e415d59784e97))

> I think you can do `go build cmd/zgrab2`, but not sure

Continuing off that, @justinbastress, could we delegate to `make clean zgrab2` or do we need to do a direct `go build`? Preference would be to keep things centralized in the `Makefile`.

docker-runner/entrypoint.sh
```diff
@@ -0,0 +1,19 @@
...
+set -e
+
+cd /go/src/github.com/zmap/zgrab2
+
+if ! [ -x $ZGRAB_REBUILD ]; then
+  if ! [ -x $ZGRAB_BRANCH ]; then
+    git checkout $ZGRAB_BRANCH
+  fi
+  git pull
+  pushd cmd/zgrab2 && go build && popd
```
andrewsardone: @justinbastress, similar question [to the above](https://github.com/zmap/zgrab2/pull/30/files#r159444127), but it'd be nice if it's possible to use `make clean zgrab2` to keep things centralized in the `Makefile` (minor nitpick, though).

docker-runner/Dockerfile
justinbastress: @dadrian indeed you can -- on Windows at least, `go build .\cmd\zgrab2` works.

@andrewsardone that's a good idea, My only concern is that it does add another dependency on the build, but `make` is probably a pretty universal pre-req; I'll confirm that the golang containers have it, and then I can start moving all of that build logic from .travis.yml into the Makefile.

integration_tests/postgres/cleanup.sh
justinbastress: In [setup.sh](https://github.com/zmap/zgrab2/blob/feature/addPostgresModule/integration_tests/postgres/setup.sh), they are launched with the `--rm` flag, so they are automatically removed when they are stopped.

justinbastress:
@andrewsardone @dadrian I should also note that having this merged is not a blocker for my upcoming tasks (e.g. porting over zgrab modules / adding new protocols).

schemas/zgrab2.py
```diff
@@ -20,7 +20,8 @@ def DebugOnly(childType):
 
 # zgrab2/processing.go: Grab
 grab_result = Record({
-    "ip": IPv4Address(required = True),
+    "ip": IPv4Address(required = False),
```
zakird: This is crazy nit-picky, but generally named params don't get spaces in Python per PEP8. So, you probably want `"ip": IPv4Address(required=False),`

justinbastress: Ran this through the PEP8 linter...should be good now.

Issue #35, opened 2018-01-15
Name Constraints schema is incorrect


dadrian:
See #34 but for name constraints.

> JSON object specified for non-record field: `data.tls.server_certificates.chain.parsed.extensions.name_constraints.excluded_ip_addresses`

dadrian:
I think it's actually just `excluded_ip_addresses` and `permitted_ip_addresses` IP addresses that are incorrect. They should be objects.

justinbastress:
Seems there were also the `edi_party_names` and `registered_ids` missing.

I did fix some blatant issues when I first copied it over, but my `openssl req -new -x509` SSL certificates obviously don't exercise the whole schema.

dadrian:
In an ideal world this is something that is tested and maintained as part of the ZCrypto project, and we generate certs with every field in the test code there.

dadrian:
Fixed in #36

Issue #34, opened 2018-01-15
Certificate Policies schema is incorrect


dadrian:
Similar to #33, but I don't think there's a script in ZCrypto to help you with it this time. It's an object, not a `ListOf(String())`.

dadrian:
Fixed in #36

Issue #33, opened 2018-01-15
Extended Key Usage is schema'd incorrectly


dadrian:
I've been poking around with the TLS parts of the schema, and noticed a few things that are off. This is all largely due to cruft, but we should make zgrab2 schema's actually match the output. I expect this "bug" is due to something we fixup in either ZTag or ESLoader.

`extended_key_usage` is typed as a `ListOf(Integer())`, but it's actually an object. There's a script in ZCrypto that helps generate the actual schema for it. See https://github.com/zmap/zcrypto/blob/6447918deb1c310126e1cecbd711bdaac3b5558d/x509/extensions.go#L47 and https://github.com/zmap/zcrypto/blob/da4419f6d9aafae65261ac8e0e4472ac99a4d284/x509/extended_key_usage_schema.sh

dadrian:
Fixed in #36

PR #36, opened 2018-01-15
Fix zcrypto schema: address issues #33, #34, #35

justinbastress:
Pull extended_key_usage, certificate_policy and name_constraints from extensions.go.


schemas/zcrypto.py
```diff
@@ -76,6 +79,105 @@
...
+ip_net = SubRecord({
+    "IP": Binary(),
+    "Mask": Binary(),
+})
+
+# x509/x509.go: type GeneralSubtreeIP struct
+general_subtree_ip = SubRecord({
+    "Data": ip_net,
+    "Max": Integer(),
+    "Min": Integer(),
```
dadrian: Shit, are these actually leading capital letter in ZCrypto?

justinbastress: I think so -- they have no `json` tag, and I don't see a custom Marshal function (https://github.com/zmap/zcrypto/blob/master/x509/x509.go#L663).

PR #42, opened 2018-01-28
Implement TLS module

dadrian:
This module just uses zgrab2.BaseTLSFlags to implement a simple TLS
module.

## How to Test

I haven't updated TravsCI or added any integration tests yet. I'm wondering if they should really be HTTP(S) tests. @justinbastress, thoughts?

## Notes & Caveats

I'm not sure what an HTTPS scan that actually fetches a page would look like under the current framework. Is `zgrab tls` going to do this, or is there some way to combine `zgrab http` and `zgrab tls`, or do we need to add a `zgrab https`? Regardless of what we choose to do with that, we'll still probably want just a raw TLS handshake mode that can be run on any port, since many protocols (that we might not implement immediately) begin with a TLS handshake (e.g. IMAPS, POP3S, etc.)

### DIFF
--- a/modules/tls.go
+++ b/modules/tls.go
@@ -7,15 +7,7 @@ import (
 
 type TLSFlags struct {
 	zgrab2.BaseFlags
-	Heartbleed           bool        `long:"heartbleed" description:"Check if server is vulnerable to Heartbleed"`
-	Version              int         `long:"version" description:"Max TLS version to use"`
-	Verbose              bool        `long:"verbose" description:"Add extra TLS information to JSON output (client hello, client KEX, key material, etc)" json:"verbose"`
-	SessionTicket        bool        `long:"session-ticket" description:"Send support for TLS Session Tickets and output ticket if presented" json:"session"`
-	ExtendedMasterSecret bool        `long:"extended-master-secret" description:"Offer RFC 7627 Extended Master Secret extension" json:"extended"`
-	ExtendedRandom       bool        `long:"extended-random" description:"Send TLS Extended Random Extension" json:"extran"`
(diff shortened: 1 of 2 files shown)


modules/tls.go
```diff
@@ -64,5 +59,17 @@ func (s *TLSScanner) InitPerSender(senderID int) error {
 }
 
 func (s *TLSScanner) Scan(t zgrab2.ScanTarget) (zgrab2.ScanStatus, interface{}, error) {
-	return zgrab2.SCAN_SUCCESS, s, nil
+	tcpConn, err := t.Open(&s.config.BaseFlags)
+	if err != nil {
+		return zgrab2.TryGetScanStatus(err), &zgrab2.TLSLog{}, err
```
justinbastress: IIRC setting result == nil is the only way to unambiguously communicate that the protocol was not detected (non-success status isn’t enough, since you could get an error collecting optional data after detecting).

But, that means you cannot return e.g. partial TLS handshakes (or even complete handshakes if there is a subsequent failure) — so we may want to revisit that.

Having said that, several other modules do the same thing, so this can probably be addressed with everything else.

justinbastress:
@dadrian by analogy with zgrab classic, I gave the http module a —https flag (or something similar); it seems if the scanner is Interested in HTTP, the transport security isn’t really relevant, while if it’s interested in TLS, the HTTP is beyond the scope. I understood TLS to do basically what you’re doing here, while HTTP (w/ the flag), SMTL, etc would get the protocol-specific results.

OTOH if it is more convenient to split them out to separate modules, there’s no reason that couldn’t work. Making the TLS smart enough to call down to another module based on e.g. the port would probably be feasible, but obviously that would require some tinkering with the interfaces.

Having said that, (re)using an HTTPS container for the TLS integration tests should be fine.

PR #43, opened 2018-01-29
zgrab2: Log ScanTarget on unrecovered panic

justinbastress:
If there is a `panic()` while scanning a target, trap it and log the target to the standard logger at the ERROR level, before re-throwing the panic.

## How to Test

There is no place place that intentionally panics, so I have been testing by adding a panic() call to e.g. `Scan@modules/ftp.go`.

## Notes & Caveats

This arose from debugging https://ichnaea.eecs.umich.edu/admin/tasks/a47e2ae2-4f47-4ae4-86da-b8cc3a85a5c9 -- since the code panicked, the IP address did not get written to the output. We don't log the IP in non-panic cases since in those cases, the IP address should be available in the output.

## Issue Tracking

Re https://trello.com/c/CcL2qo60/1191-zgrab-https-scan-segfault, which was fixed by https://github.com/zmap/zcrypto/pull/107.

### DIFF
--- a/processing.go
+++ b/processing.go
@@ -26,6 +26,18 @@ type ScanTarget struct {
 	Domain string
 }
 
+func (self ScanTarget) String() string {
+	if self.IP == nil && self.Domain == "" {
+		return "<empty target>"
+	} else if self.IP != nil && self.Domain != "" {
+		return self.Domain + "(" + self.IP.String() + ")"
+	} else if self.IP != nil {
+		return self.IP.String()
+	} else {
+		return self.Domain
+	}
+}
+
 // scanTargetConnection wraps an existing net.Conn connection, overriding the Read/Write methods to use the configured timeouts
 type scanTargetConnection struct {
 	net.Conn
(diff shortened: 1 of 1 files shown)


processing.go
```diff
@@ -26,6 +26,18 @@ type ScanTarget struct {
 	Domain string
 }
 
+func (self ScanTarget) String() string {
```
dadrian: `self`? What is this, Python? 😝

processing.go
```diff
@@ -78,6 +90,13 @@ func grabTarget(input ScanTarget, m *Monitor) []byte {
 	moduleResult := make(map[string]ScanResponse)
 
 	for _, scannerName := range orderedScanners {
+		defer func(name string) {
+			if e := recover(); e != nil {
+				log.Errorf("Panic on scanner %s when scanning target %s", scannerName, input.String())
+				// Bubble out original error (with original stack) in lieu of explicitly logging the stack / error
+				panic(e)
```
dadrian: So we actually get the original stack trace here, we just get to shove in a log statement? That's awesome!

justinbastress: I was suspicious too, so I tried it out, and it works as advertised.

PR #47, opened 2018-01-31
Fix README install instructions pointing to previous versions of zgrab

jordan-wright:
This is a small change to the `README`. I'm honestly not sure if this is incorrect documentation or not since _technically_ zgrab2 has zgrab as a dependency so you _can_ go into the old zmap directory and build it.

But my guess was that y'all meant to reference the new version?

## How to Test

No changes

## Notes & Caveats

None

## Issue Tracking

Didn't think this really deserved an issue 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,10 +13,10 @@ Once you have a working `$GOPATH`, run:
 $ go get github.com/zmap/zgrab2
 ```
 
-This will install zgrab under `$GOPATH/src/github.com/zmap/zgrab`
+This will install zgrab under `$GOPATH/src/github.com/zmap/zgrab2`
 
 ```
-$ cd $GOPATH/src/github.com/zmap/zgrab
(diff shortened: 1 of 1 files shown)


dadrian:
We have ZGrab as a dependency?

jordan-wright:
Looks like it's probably when importing the ztoolsKeys library? https://github.com/zmap/zgrab2/search?utf8=%E2%9C%93&q=ztoolsKeys&type=

ajholland:
🤦‍♂️  Looks like the ZGrab as a dependency was me. Looks like that could just be switched s#zgrab/ztools#zgrab2/tools#g as ztoolsKeys is already in this repo.

dadrian:
@ajholland Can you file an issue?

jordan-wright:
FWIW, I'm already working on a patch for that, too 😄 Just checking the integration tests and whatnot - should have a pull submitted in a couple of hours.

PR #49, opened 2018-01-31
Changed tools/keys dependencies to point to zgrab2.

### LINKED ISSUE #48: Remove ZGrab as a dependency
s#zgrab/ztools#zgrab2/tools#g

https://github.com/zmap/zgrab2/search?utf8=%E2%9C%93&q=ztoolsKeys&type=

jordan-wright:
This is the first step towards removing the dependency on zgrab.

## How to Test

Same way as always.

## Notes & Caveats

*Note: I'm sure y'all are on it without me mentioning it, but we should wait until Travis runs the integration tests here... I was having issues getting the integration tests to complete successfully even from a clean build (I'll make an issue for that).*

Unfortunately, even without these changes there will still be nested dependencies on the original `zgrab/ztools/keys` package through `github.com/zmap/zcrypto` https://godoc.org/github.com/zmap/zgrab/ztools/keys?importers

Changes will be needed to zcrypto to fully remove the dependency on zgrab.

## Issue Tracking

Doesn't fully fix #48, but this at least starts to address it.

### DIFF
--- a/lib/ssh/kex.go
+++ b/lib/ssh/kex.go
@@ -17,7 +17,7 @@ import (
 	"io"
 	"math/big"
 
-	ztoolsKeys "github.com/zmap/zgrab/ztools/keys"
+	ztoolsKeys "github.com/zmap/zgrab2/tools/keys"
 
 	"golang.org/x/crypto/curve25519"
 )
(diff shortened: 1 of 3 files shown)


justinbastress:
@jordan-wright nice catch; I thought I had replaced all of these, but apparently those were missed for some reason.

PR #53, opened 2018-02-06
Add redis zgrab2 module

justinbastress:
Added a basic redis module.

Pushed yet more of the boilerplate into the new module setup script.

## How to Test

As always, `make integration-test` will work. Or, to test manually:

* `docker run -td -p 6379:6379 --name zredis --rm redis`
* `echo "localhost" | ./zgrab2/cmd/zgrab2 redis`

You can force authentication by running
* `nc localhost 6379`
* `requirepass your-password`

## Issue Tracking

* Tracked in Trello.


modules/redis/scanner.go
```diff
@@ -0,0 +1,254 @@
...
+type Module struct {
+}
+
+// Scanner implements the zgrab2.Scanner interface
+type Scanner struct {
+	config *Flags
+}
+
+// Scan holds the state for the scan of an individual target
+type Scan struct {
```
dadrian: It looks like a lot of modules have a `Scan` struct that somehow gets converted into a `Result` struct. Is this something that we should consider at the framework level? Should these be public or private?

modules/redis/scanner.go
```diff
@@ -0,0 +1,254 @@
...
+	InfoResponse string `json:"info_response,omitempty"`
+
+	// QuitResponse is the response from the QUIT command -- should be the
+	// simple string "OK" even when authentication is required, unless the
+	// QUIT command was renamed.
+	QuitResponse string `json:"quit_response,omitempty"`
+
+	// NonexistentResponse is the response to the non-existent command; even if
+	// auth is required, this may give a different error than existing commands.
+	NonexistentResponse string `json:"nonexistent_response,omitempty"`
```
dadrian: Is NONEXISTANT an actual command? Or are we testing the error response of a nonexistant command?

modules/redis/types.go
```diff
@@ -0,0 +1,422 @@
...
+
+// Encode returns the encoding of the error message ("-<error message>\r\n")
+func (err ErrorMessage) Encode() []byte {
+	return []byte("-" + err + "\r\n")
+}
+
+// ErrorPrefix returns the first word of the error message, which can be
+// interpreted as a sort of error code.
+func (err ErrorMessage) ErrorPrefix() string {
+	return strings.SplitN(string(err), " ", 2)[0]
```
dadrian: Do we need a `len(err) > 0` check here?

modules/redis/types.go
```diff
@@ -0,0 +1,422 @@
...
+	return string(err)
+}
+
+// Integer type -- an int64; "the returned integer is guaranteed to be in the
+// range of a signed 64 bit integer"
+// See https://redis.io/topics/protocol#resp-integers
+type Integer int64
+
+// Type identifies this instance as a TypeInteger
+func (Integer) Type() RedisType {
```
dadrian: I didn't know you can have anonymous "this" objects!

modules/redis/scanner.go
```diff
@@ -0,0 +1,254 @@
...
+// If authentication is required, most responses can have the value
+// "(error: NOAUTH Authentication required.)"
+type Result struct {
+	// Commands is the list of commands actually sent to the server, serialized
+	// in inline format (e.g. COMMAND arg1 "arg 2" arg3)
+	Commands []string `json:"commands,omitempty" zgrab:"debug"`
+
+	// CommandOutput is the output returned by the server for each command sent;
+	// the index in CommandOutput matches the index in Commands.
+	CommandOutput [][]byte `json:"command_output,omitempty" zgrab:"debug"`
```
dadrian: We usually put raw in the front of unparsed binary outputs to make it more clear, e.g. `raw_command_output`.

modules/redis/scanner.go
justinbastress: I usually use a `Scan` if I need to keep track of state for one specific scan -- basically, if the scan gets too big to have everything in local variables in the `Scanner.Scan` method.
There would by definition be a 1:1 relationship between the `Scan` instances and results (just like there would be between `ScanTarget` and results), but I don't necessarily see it as a conversion.

Continuing the "encapsulated local variables" analogy, it would make sense for the `Scan` struct to be private. Anything outside of the module should be using the Scanner interface.

modules/redis/scanner.go
justinbastress: The latter -- though even if it does for some reason exist, it is still the response to the command "NONEXISTENT" in the same sense that QuitResponse is the response to the command "QUIT".

modules/redis/types.go
justinbastress: Confirmed that `len(strings.SplitN("", " ", 2)) == 1`, but it would probably be good to make it explicit.

PR #55, opened 2018-02-09
Run golint on postgres module

justinbastress:
Mostly comment / godoc changes. 

Important changes -- package godocs, flags/results godocs, detailed Scan godocs.

Code changes -- a few `if (x) { return a } else { return b }` replaced by `if (x) { return a };  return b` and a couple of `log.Fatalf()` format string issues.

## How to Test

Run integration tests and/or view godocs.

## Notes & Caveats

The hope is that the scanner documentation can just point to the godocs.

### DIFF
--- a/modules/postgres/connection.go
+++ b/modules/postgres/connection.go
@@ -14,31 +14,42 @@ import (
 	"github.com/zmap/zgrab2"
 )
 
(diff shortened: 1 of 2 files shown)


zakird:
looks good to merge once tests pass ;)

PR #60, opened 2018-02-12
Update MSSQL GoDocs

justinbastress:
* re-export MSSQL primitives, remove stuttering `mssql.mssqlXYZ`  names
 * re-run golint / gofmt
 * add package godocs
 * add detailed scan docs
 * bugfixes:
  * populate InstanceName in result
  * fix return value in case of failed / incomplete detection
  * add default EncryptMode

## How to Test

`TEST_MODULES=mssql make integration-test`

 * Confirm that InstanceName is set
 * Point to a non-mssql service, like HTTP, and confirm that it returns a nil result

### DIFF
--- a/modules/mssql/connection.go
+++ b/modules/mssql/connection.go
@@ -4,6 +4,7 @@ import (
 	"encoding/binary"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io"
 	"net"
 	"sort"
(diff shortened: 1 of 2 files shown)


modules/mssql/connection.go
```diff
@@ -15,97 +16,197 @@ import (
...
+
+	// TDSPacketTypeRPC identifies an RPC packet.
+	TDSPacketTypeRPC = 0x03
+
+	// TDSPacketTypeTabularResult identifies a tabular result.
+	TDSPacketTypeTabularResult = 0x04
+
+	// TDSPacketTypeAttentionSignal identifies an attention signal.
+	// Packet does not contain data.
+	TDSPacketTypeAttentionSignal = 0x06
```
dadrian: This whole section is real nice.

PR #64, opened 2018-02-28
Make travis run unit tests

justinbastress:
* Added phony `test` Makefile target
 * Make travis run the `test` target before getting into the much longer integration tests

## How to Test

Run `make test`, or kick off a travisci build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ before_install:
 - pip install --user zschema
 - docker ps -a
 install:
-- make clean zgrab2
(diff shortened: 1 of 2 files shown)


zakird:
👍🏽 merge when tests run?

justinbastress:
I think what keeps happening is, the PR and Push CI tasks run simultaneously, so one of them gets the short end of the stick and ends up timing out.

Then when I re-run the loser on its own, it succeeds.

PR #68, opened 2018-03-07
Custom HTTP Headers WIP

zakird:
This is a work in progress. I haven't had a chance to test yet, but I was curious what ZGrab 2.0 looked like these days, so I decided to try and tackle.

### DIFF
--- a/modules/http/scanner.go
+++ b/modules/http/scanner.go
@@ -9,11 +9,13 @@ package http
 import (
 	"bytes"
 	"crypto/sha256"
+	"encoding/csv"
 	"errors"
 	"io"
 	"net"
 	"net/url"
 	"strconv"
+	"strings"
 	"time"
 
 	log "github.com/sirupsen/logrus"
@@ -50,7 +52,9 @@ type Flags struct {
 
 	// UseHTTPS causes the first request to be over TLS, without requiring a
 	// redirect to HTTPS. It does not change the port used for the connection.
-	UseHTTPS bool `long:"use-https" description:"Perform an HTTPS connection on the initial host"`
+	UseHTTPS            bool   `long:"use-https" description:"Perform an HTTPS connection on the initial host"`
+	CustomHeadersNames  string `long:"custom-headers-names" description:"CSV of custom HTTP headers to send to server"`
+	CustomHeadersValues string `long:"custom-headers-values" description:"CSV of custom HTTP header values to send to server. Should match order of custom-headers-names."`
 }
 
 // A Results object is returned by the HTTP module's Scanner.Scan()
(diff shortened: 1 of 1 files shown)


modules/http/scanner.go
```diff
@@ -245,6 +283,11 @@ func (scan *scan) Grab() *zgrab2.ScanError {
 	}
 	// TODO: Headers from input?
 	request.Header.Set("Accept", "*/*")
+	if scan.scanner.customHeaders != nil {
```
justinbastress: If custom headers are provided, should that list be the *only* headers that are sent? The way you have it implemented is probably the most intuitive, but at the cost of flexibility -- e.g. as it stands you can change the Accept header, but you cannot remove it.

zakird: Good point, I'll fix it to match your recommendation.

mzpqnxow:
@zakird , @justinbastress , @corny I just opened #272 before realizing this effort had already begun. Is there any movement on this or is it on hold? I'm happy to pitch in here either by testing what's there or changing to fit your requirements/suggestions

zakird:
No one is actively working on it. If you have time, we'd love help on it.

modules/http/scanner.go
mzpqnxow: > If custom headers are provided, should that list be the _only_ headers that are sent? The way you have it implemented is probably the most intuitive, but at the cost of flexibility -- e.g. as it stands you can change the Accept header, but you cannot remove it.

I just noticed that if you separate this into two cases- one for when the user is specifying custom headers, one for where they're not- and have the custom header case *not* set the `Accept` header here- it just gets set later on in the HTTP library to a default value of `gzip`, which is probably not what you want

The code for that is  [here](https://github.com/zmap/zgrab2/blob/master/lib/http/transport.go#L1869) 

So maybe it is best to always assign `*/*` to `Accept` by default, allow the custom headers to override the value, and then if you want it to be removable, use a `--no-accept-header` flag? The only other option is to modify code in the HTTP library linked above. After noticing this, I'm inclined to leave it as is, but am curious what you think

zakird: > So maybe it is best to always assign */* to Accept by default, allow the custom headers to override the value, and then if you want it to be removable, use a --no-accept-header flag?

It sounds like this is the simplest, if I understand correctly. If so, that sounds like a good option to me.

mzpqnxow:
@zakird I started to work on this and then realized after testing that the existing PR works pretty well as is. Are there any specific guidelines or feedback as to why it’s not an acceptable solution? I saw some comments but if I’m reading correctly it seems like the recommended changes were implemented. If a change to this is required I think I’ll need a little guidance on what needs changing (at a high level)

Or maybe it can be merged as-is?

zakird:
I think I just never finished it. I'm not sure that it works.

mzpqnxow:
It definitely works, so you’re all set in that regard.

I’ll PR the version I used, which was your code shoehorned into the current master and then we can go from there. I can include some test data with it

mzpqnxow:
FYI @zakird this one can be closed, the rebased version is in #272

PR #75, opened 2018-03-14
zgrab2: Port SMTP module from ZGrab

justinbastress:
Behavior should be identical to the original zgrab.

## How to Test

`echo "smtp.gmail.com" | cmd/zgrab2/zgrab2 smtp --timeout 3 --ehlo <your IP address> --starttls`

## Notes & Caveats

No unit tests (none in original); no integration tests (I was able to manually set up an opensmtpd container, but doing it automatically would take some additional work)

## Issue Tracking

Jira issue CEN-73


zakird:
It might make sense for `--ehlo` to be default enabled.

schemas/smtp.py
```diff
@@ -0,0 +1,22 @@
...
+import zschema.registry
+
+import schemas.zcrypto as zcrypto
+import schemas.zgrab2 as zgrab2
+
+smtp_scan_response = SubRecord({
+    "result": SubRecord({
+        "banner": String(),
+        "ehlo": String(),
+        "smtp_help": String(),
```
zakird: Seems like this could just be `help`. Should be apparent that it's for SMTP...

modules/smtp/scanner.go
```diff
@@ -0,0 +1,199 @@
...
+// ScanResults instances are returned by the module's Scan function.
+type ScanResults struct {
+	// Banner is the string sent by the server immediately after connecting.
+	Banner string `json:"banner,omitempty"`
+
+	// EHLO is the server's response to the EHLO command, if one is sent.
+	EHLO string `json:"ehlo,omitempty"`
+
+	// SMTPHelp is the server's response to the HELP command, if it is sent.
+	SMTPHelp string `json:"smtp_help,omitempty"`
```
zakird: As noted below too, probably fine to shorten this to just `help`

schemas/smtp.py
justinbastress: I can get behind that. I originally left it there for symmetry with the input `—smtp-help` (since `—help` is taken), but that’s not a concern on the output.

zakird: I figured that was the reason. I wondered if something like `--help-command` or `--send-help` might be less confusing than `--smtp-help`, since that might also sound like something that provides the user with help information.

justinbastress:
You can do HELP and QUIT without EHLO/HELO, but not STARTTLS (at least, not in the servers I tried), so omitting it (and `—help`) would be the least-invasive scan (making QUIT optional would go further in that direction).

If EHLO were on by default, what would the host be? I can check if it can be sent without one (like IIRC HELO could), but that might be a way to handle the TODO regarding using the scanner IP.

justinbastress:
@zakird I think I addressed most of your comments: `--smtp-help` is now `--send-help`, alongside `--send-ehlo`, `--send-helo`, `--send-quit`. The output for the `--send-wxyz` is now `wxyz`.

I left EHLO off by default, and made QUIT off by default as well, so now the default scan will in fact be silent from the client side, and it will just read the banner.

If `--ehlo-domain` is left out, I have it just send `EHLO` with no domain -- which "SHOULD"
 not be done (https://tools.ietf.org/html/rfc2821#section-4.1.1.1), but nonetheless can give a result useful for distinguishing different server software. Doing the scanner IP still might be the better solution, but oh well.

PR #79, opened 2018-03-15
zgrab2: Port SMB scanner from ZGrab

justinbastress:
Copied the SMB behavior from ZGrab -- namely, copied the hacked/pseudo-vendored SMB lib from zgrab/ztools/smb into zgrab2, and called its GetBanner method.

*BUT*, I have yet to have a positive hit with this (though I get the same results with the original zgrab, so maybe both my container Samba and Windows NetBIOS versions are too new).

## How to Test

On a host that supports SMBv1, run `echo "that-host" | cmd/zgrab2/zgrab2 smb`.

## Notes & Caveats

If I pull in the latest version of the SMB library (github.com/stacktitan/smb), it succeeds in detecting my test machines -- but, we would need to re-apply our hacks to give ourselves lower-level access.

## Issue Tracking

Jira issue CEN-69.

### DIFF
--- a/lib/smb/encoder/encoder.go
+++ b/lib/smb/encoder/encoder.go
@@ -1,0 +1,490 @@
+/*
+* MIT License
+*
+* Copyright (c) 2017 stacktitan
+*
+* Permission is hereby granted, free of charge, to any person obtaining a copy
+* of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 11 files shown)


zakird:
I'm wondering if we should wait to pull this in if this doesn't find machines but the upstream library does. This might be a good opportunity to fix something broken in our pipeline.

PR #88, opened 2018-03-22
allow users to specify a comma-delimited list of ciphers to send

zakird:
This allows users to specify the cipher suites they want to sen.

## How to Test

```echo "google.com" | ./zgrab2 tls --cipher-suite=0xC02F,0xC02F``` and look at the cipher suites we send.

### DIFF
--- a/tls.go
+++ b/tls.go
@@ -5,6 +5,7 @@ import (
 	"encoding/csv"
 	"fmt"
 	"net"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: tls.go [hunk 1]
```diff
@@ -137,9 +138,21 @@ func (t *TLSFlags) GetTLSConfig() (*tls.Config, error) {
...
-			return nil, fmt.Errorf("%s is not a valid cipher suite", t.CipherSuite)
+		// allow either one of our standard values (e.g., chrome) or a comma-delimited list of ciphers
+		if _, ok := cipherMap[t.CipherSuite]; ok {
+			ret.CipherSuites = cipherMap[t.CipherSuite]
+		} else {
+			strCiphers := getCSV(t.CipherSuite)
+			var intCiphers = make([]uint16, len(strCiphers))
+			for i, s := range strCiphers {
+				s = strings.TrimPrefix(s, "0x")
+				v64, err := strconv.ParseUint(s, 16, 16)
```
justinbastress: Here you *could* use radix = 0 to interpret the 0x automatically (and if they decide they want to put it in decimal or octal, that would work too). But then, that may not be desirable in this case.

Either way, this should probably just go into zflags -- []typeX just runs the typeX parser on each comma-separated element.

Issue #29, opened 2017-12-21
zgrab2 JSON Marshaller


justinbastress:
zgrab2 needs a special JSON Marshaller that can take into account command line flags and tags like zgrab.debug

PR #93, opened 2018-03-23
zgrab2: Modified custom JSON marshaler

justinbastress:
Added a modified copy of the golang `json/encoding` to `lib/json`, whose `Marshal()` takes a `debug` boolean -- and if this boolean is false, fields tagged with `zgrab:"debug"` are omitted from the output.

## How to Test

`make integration-test` -- then check zgrab-output. If `./zgrab-output/mysql/8.0.json` still has `raw_packets`, it's not working.

## Notes & Caveats

See the README.md (https://github.com/zmap/zgrab2/blob/71e386d86dc0e761ec06c111289b19dc7071d66a/lib/json/README.md)

## Issue Tracking

Issue #29


zakird:
Uff, this one is long for review 🙃. Will leave to @dadrian to confirm this approach is solid.

dadrian:
I'm pretty against forking yet another Golang standard library. @justinbastress, would you be opposed to trying to write a version of that uses reflection and seeing how ugly/bad/impossible/doable it is?

I could see a few options. The ideal being `byteOutput := ZGrabMarshal(&s, VerbosityDebug)` which is a pure function that delegates to `json.Marshal()` and does what we want. Internally, it might copy the struct and `nil` out all fields that are at too high of a verbosity level.

If that ends up being complicated, the other thing I could see is 
```
interfaceOutput := ZGrabNilVerbosityAbove(&s, VerbosityInfo)
byteOutput := json.Marshal(interfaceOutput)
```
This returns a new object implementing `interface{}` that has the necessary JSON tags, but with niled fields that shouldn't be output, and then sends that to JSON. Worst case scenario, this could be a non-pure function that just modifies the input struct, but the less mutations the better, IMHO.

justinbastress:
That should work. Walking the structure and making a deep copy (omitting fields depending on the struct’s tags) should be feasible; types with custom `MarshalJSON` methods will still be problematic, but that goes for the forked implementation, too.

PR #97, opened 2018-04-03
zgrab2: Strip debug fields

justinbastress:
Add `output.Process(scanResults)` and `output.NewProcessor().Process(scanResults)`, which returns a copy of `scanResults` with debug-only fields nilled out.

## How to Test

Any TLS scan should now have `client_hello` removed. `make integration-test` and check zgrab-output, e.g. `mysql/8.0.json`, `ftp/authtls.json`, etc.

`make test` now includes `cd lib/output/test && go test -v`, which exercises the `Process()` methods with both verbose values on many types of output.

## Notes & Caveats

Anything with pointers to hidden fields and/or custom JSON processors may not work as expected, but the scan results structs ideally wouldn't have a lot of that anyway.

## Issue Tracking

Jira CEN-223.


justinbastress:
@dadrian My thinking with the name `output.Process()` was that any other output processing could happen here as well; perhaps ztag-like transforms or the like.

Issue #73, opened 2018-03-13
Change HTTP timeout to be a float


ngngng:
Hi,
I wanted to set the timeout in http to a value less than a second
However the minimum allowed value is 1 second (I wanted for example 0.2 second, 200 milliseconds)
Thanks!

zakird:
@ngngng We'd be happy to see this! Would you be willing to submit a pull request that makes the change?

phillip-stephens:
Looks like this was addressed in #111 

Can confirm in latest `master` that this works and sets the timeout at 100 ms.
```shell
echo "prstephens.com" | ./zgrab2 http --timeout=0.1s --max-redirects=1
...
:"net/http: request canceled (Client.Timeout exceeded while awaiting headers)"}}}
INFO[0000] finished grab at 2025-02-04T18:42:04Z
{"statuses":{"http":{"successes":0,"failures":1}},"start":"2025-02-04T18:42:04Z","end":"2025-02-04T18:42:04Z","duration":"106.215443ms"}
```

PR #111, opened 2018-05-03
Implement timeout to accept float value

ajholland:
This PR implements changing timeout from an int value to a float so that one may use 0.5 as a timeout for example. 

## How to Test

`echo httpstat.us | zgrab2 http --timeout 3.5 --endpoint="/200?sleep=7000"`

## Notes & Caveats

I'm not sure why lines 81-86 in modules/http/scanner.go appear as different, those came from go fmt running when I saved the file. I'm guessing this is because I'm using go version 1.9. I can change that if you'd like.

Tacking on a "s" in order to use ParseDuration isn't ideal, but is in my opinion more simple than writing a custom function to parse out seconds, milliseconds, etc...

## Issue Tracking

#73


justinbastress:
Since it's a string anyway, any reason we shouldn't just let the user specify the suffix, rather than unconditionally tacking on the `"s"`? If the value ends in a number we can still append the "s", but otherwise it could be nice to allow e.g "10m", "0.5s", "500us", etc.

justinbastress:
Actually, it would be nice to have this in zflags -- i.e., specifying the input type as `time.Duration`, it does this conversion automatically.

ajholland:
Apparently, it does this already!

justinbastress:
I'm ambivalent as to requiring the unit suffix, but that would obviously be a zflags feature, and I see no problem with explicitly specifying it in the integration scripts here.

zakird:
@ajholland I can't tell from the examples, right now, if you just pass in a straight integer, e.g., `10` not `10s`, will it treat that as seconds?

zakird:
I've fixed the merge conflict, but it'd be nice to get this merged before we have to do that again. ;)

ajholland:
screen shot 2018-06-28 at 11 02 09 pm
This is the behavior in this PR for passing in 10 vs 10s. 

I could write a function (probably in zflags) that checks if units are given, if not, assumes seconds?

zakird:
I think that would probably be the most user-friendly option if it's not awful to implement.

ajholland:
@zakird @justinbastress I think this can be merged now

Issue #115, opened 2018-05-10
mssql: Strip trailing null terminator from instance name


justinbastress:
Empty instance names are turning up as `"\u0000"`. These should be just empty.

justinbastress:
Fixed.

Issue #114, opened 2018-05-10
mysql: Disable debug logs in non-verbose mode


justinbastress:
Logs are getting spammed with e.g.

[...]

PR #120, opened 2018-05-21
Address zgrab2 issues

justinbastress:
* Add `error_id` to mysql
* Remove trailing null terminator from mssql instance_name
* Catch `ERRPacket` in mysql

## How to Test

`make integration-test`

## Notes & Caveats

`generate_mysql_errors()` is mostly for reference; these shouldn't change frequently.

## Issue Tracking

#114, #115


lib/mysql/mysql.go
```diff
@@ -693,8 +715,8 @@ func (c *Connection) Connect(conn net.Conn) error {
 		c.ConnectionLog.Handshake = packet
 	case *ERRPacket:
 		c.ConnectionLog.Error = packet
-		log.Debugf("Got error packet: 0x%x / %s", p.ErrorCode, p.ErrorMessage)
-		return fmt.Errorf("Server returned error after connecting: error_code = 0x%x; error_message = %s", p.ErrorCode, p.ErrorMessage)
+		log.Debugf("Got error packet: %s", p.Error())
+		return p.GetScanError()
```
justinbastress: Before this, it giving the `SCAN_UNKNOWN_ERROR` status. Returning the `ScanStatus` allows it to detect the proper `SCAN_APPLICATION_ERROR`.

zgrab2_schemas/zgrab2/mysql.py
```diff
@@ -60,20 +60,79 @@
...
         "error_code": Signed32BitInteger(doc="Only set if there is an error returned by the server, for example if the scanner is not on the allowed hosts list."),
+        # error_ids could be an Enum(values=mysql_errors.mysql_error_code_to_id), but that would be brittle to changes.
+        "error_id": String(doc="The friendly name for the error code as defined at https://dev.mysql.com/doc/refman/8.0/en/error-messages-server.html, or UNKNOWN."),
         "error_message": String(doc="Optional string describing the error. Only set if there is an error."),
         "raw_packets": ListOf(Binary(), doc="The base64 encoding of all packets sent and received during the scan."),
         "tls": zgrab2.tls_log,
     })
 }, extends=zgrab2.base_scan_response)
 
+def generate_mysql_errors(root='.'):
```
justinbastress: This is what generated `mysql_errors.py` and `lib/mysql/errors.go`.  It probably won't need to be run again, but it's here for tracking purposes if nothing else.

PR #135, opened 2018-06-08
Scaffold for IPP scan module

clayzim:
Basic implementation of IPP scanning.

## How to Test

`make integration-test` from zgrab2 root directory.

## Notes & Caveats
Overall, could use adjustments to ScanResult items and some refactoring.
Tests are not yet fully implemented. Currently, they pass regardless of whether the correct version is returned, though the result is correct.

## Issue Tracking

N/A

### DIFF
--- a/integration_tests/ipp/container/Dockerfile
+++ b/integration_tests/ipp/container/Dockerfile
@@ -1,16 +1,17 @@
 FROM zgrab2_service_base:latest
 
-RUN apt-get update
-RUN apt-get install -y cups
-RUN apt-get install -y cups-pdf
-RUN echo "ServerAlias *" >> /etc/cups/cupsd.conf
-RUN echo "FileDevice yes" >> /etc/cups/cupsd.conf
-RUN sed -i 's/Order allow,deny/Order allow,deny\'$'\n  Allow all/g' /etc/cups/cupsd.conf
-RUN sed -i 's/Listen localhost:631/Listen 0.0.0.0:631/g' /etc/cups/cupsd.conf
+RUN apt-get update && apt-get install -y \
+  cups \
+  cups-pdf
+# TODO: Provide a pre-built cupsd.conf rather than relying on modifying the default config file
+WORKDIR /etc/cups
+COPY cupsd.conf cupsd.conf
+# TODO: Provide a pre-built cups-pdf.conf
 RUN service cups restart
 
-# Try to make it act more container-y -- remove it from init.d and just run the daemon as the entrypoint
+# TODO: Actually stop service; see why this works without stopping it
 #RUN service cupsd stop
+# TODO: Actually prevent cups from being started automatically
 #RUN update-rc.d -f cupsd remove
 
 WORKDIR /
--- a/integration_tests/ipp/container/Dockerfile
+++ b/integration_tests/ipp/container/Dockerfile
@@ -5,10 +5,9 @@ RUN apt-get install -y cups
 RUN apt-get install -y cups-pdf
 RUN echo "ServerAlias *" >> /etc/cups/cupsd.conf
 RUN echo "FileDevice yes" >> /etc/cups/cupsd.conf
-#FIXME: Hard-coded IP address of my machine, which totally doesn't work for real testing
 RUN sed -i 's/Order allow,deny/Order allow,deny\'$'\n  Allow all/g' /etc/cups/cupsd.conf
-#RUN sed -i 's/Listen localhost:631/Listen 0.0.0.0:631/g' /etc/cups/cupsd.conf
-RUN service cupsd restart
+RUN sed -i 's/Listen localhost:631/Listen 0.0.0.0:631/g' /etc/cups/cupsd.conf
+RUN service cups restart
 
 # Try to make it act more container-y -- remove it from init.d and just run the daemon as the entrypoint
 #RUN service cupsd stop
@@ -18,4 +17,4 @@ WORKDIR /
 COPY entrypoint.sh .
 RUN chmod a+x ./entrypoint.sh
 
-#ENTRYPOINT ["/entrypoint.sh"]
+ENTRYPOINT ["/entrypoint.sh"]
--- a/integration_tests/ipp/container/Dockerfile
+++ b/integration_tests/ipp/container/Dockerfile
@@ -1,10 +1,18 @@
 FROM zgrab2_service_base:latest
 
+RUN apt-get update
 RUN apt-get install -y cups
+RUN apt-get install -y cups-pdf
+RUN echo "ServerAlias *" >> /etc/cups/cupsd.conf
+RUN echo "FileDevice yes" >> /etc/cups/cupsd.conf
+#FIXME: Hard-coded IP address of my machine, which totally doesn't work for real testing
+RUN sed -i 's/Order allow,deny/Order allow,deny\'$'\n  Allow all/g' /etc/cups/cupsd.conf
+#RUN sed -i 's/Listen localhost:631/Listen 0.0.0.0:631/g' /etc/cups/cupsd.conf
+RUN service cupsd restart
 
 # Try to make it act more container-y -- remove it from init.d and just run the daemon as the entrypoint
-#RUN service FIXME_YOUR_SERVICED stop
-#RUN update-rc.d -f FIXME_YOUR_SERVICED remove
+#RUN service cupsd stop
+#RUN update-rc.d -f cupsd remove
 
 WORKDIR /
 COPY entrypoint.sh .
--- a/integration_tests/ipp/container/Dockerfile
+++ b/integration_tests/ipp/container/Dockerfile
@@ -1,0 +1,13 @@
+FROM zgrab2_service_base:latest
+
+RUN apt-get install -y cups
+
+# Try to make it act more container-y -- remove it from init.d and just run the daemon as the entrypoint
+#RUN service FIXME_YOUR_SERVICED stop
+#RUN update-rc.d -f FIXME_YOUR_SERVICED remove
+
+WORKDIR /
+COPY entrypoint.sh .
+RUN chmod a+x ./entrypoint.sh
+
+#ENTRYPOINT ["/entrypoint.sh"]
--- a/modules/ipp/ipp.go
+++ b/modules/ipp/ipp.go
@@ -1,0 +1,70 @@
+package ipp
+
+import (
+	"bytes"
+	"encoding/binary"
+	//"io"
+	"net"
+)
+
+type Connection struct {
+	Conn net.Conn
+}
+
+//func ReadResponse(body *io.ReadCloser) *ScanResults {
+//	result := &ScanResults{}
+//
+//}
+
+func reverse(b []byte) []byte {
+	for i := 0; i < len(b)/2; i++ {
+		b[i], b[len(b)-1-i] = b[len(b)-1-i], b[i]
+	}
+	return b
+}
+
+//FIXME: Clean this up, and use binary package to handle endianness in the correct order
+//FIXME: Make sure uint isn't messing anything up here, since they should be signed.
+func attributeByteString(syntaxTag byte, name string, value string) []byte {
+	//special bytestring denoting value syntax
+	b := []byte{syntaxTag}
+	l := make([]byte, 2)
+	binary.PutUvarint(l, uint64(len(name)))
+	l = reverse(l)
+	b = append(b, l...)
+	b = append(b, []byte(name)...)
+
+	l = make([]byte, 2)
+	binary.PutUvarint(l, uint64(len(value)))
+	l = reverse(l)
+	b = append(b, l...)
+	b = append(b, []byte(value)...)
+	return b
+}
+
+//Construct a minimal request that an IPP server will respond to
+func getPrinterAttributesRequest(uri string) bytes.Buffer {
+	var b bytes.Buffer
+	//version 2.1 (newest as of 2018)
+	b.Write([]byte{2, 1})
+	//operation-id = get-printer-attributes
+	b.Write([]byte{0, 0xb})
+	//request-id = 1
+	b.Write([]byte{0, 0, 0, 1})
+	//operation-attributes-tag = 1
+	b.Write([]byte{1})
+
+	//attributes-charset
+	b.Write(attributeByteString(0x47, "attributes-charset", "utf-8"))
+	//attributes-natural-language
+	b.Write(attributeByteString(0x48, "attributes-natural-language", "en-us"))
+	//printer-uri
+	b.Write(attributeByteString(0x45, "printer-uri", uri))
+	//requested-attributes
+	b.Write(attributeByteString(0x44, "requested-attributes", "all"))
+
+	//end-of-attributes-tag = 3
+	b.Write([]byte{3})
+
+	return b
+}
--- a/modules/ipp/ipp.go
+++ b/modules/ipp/ipp.go
@@ -16,33 +16,38 @@ type Connection struct {
 //
 //}
 
-func reverse(b []byte) []byte {
-	for i := 0; i < len(b)/2; i++ {
-		b[i], b[len(b)-1-i] = b[len(b)-1-i], b[i]
-	}
-	return b
-}
+// Returns a byte-encoded "attribute-with-one-value" with the provided "value-tag", "name", and "value"
+// attribute-with-one-value encoding described at https://tools.ietf.org/html/rfc8010#section-3.1.4
+// Example (runnable from ipp_test.go):
+//   Input: 0x47, "attributes-charset", "us-ascii"
+//   Output: [71 0 18 97 116 116 114 105 98 117 116 101 115 45 99 104 97 114 115 101 116 0 8 117 115 45 97 115 99 105 105]
+// TODO: Should return an error when fed an invalid valueTag?
+// TODO: Determine whether this should remain public. Currently is for Testable Example
+func AttributeByteString(valueTag byte, name string, value string) []byte {
+	//special byte denoting value syntax
+	b := []byte{valueTag}
 
-//FIXME: Clean this up, and use binary package to handle endianness in the correct order
-//FIXME: Make sure uint isn't messing anything up here, since they should be signed.
-func attributeByteString(syntaxTag byte, name string, value string) []byte {
-	//special bytestring denoting value syntax
-	b := []byte{syntaxTag}
-	l := make([]byte, 2)
-	binary.PutUvarint(l, uint64(len(name)))
-	l = reverse(l)
-	b = append(b, l...)
+	//append 16-bit signed int denoting name length
+	l := new(bytes.Buffer)
+	binary.Write(l, binary.BigEndian, int16(len(name)))
+	b = append(b, l.Bytes()...)
+
+	//append name
 	b = append(b, []byte(name)...)
 
-	l = make([]byte, 2)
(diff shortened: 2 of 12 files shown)


integration_tests/ipp/container/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM zgrab2_service_base:latest
+
+RUN apt-get update
+RUN apt-get install -y cups
+RUN apt-get install -y cups-pdf
+RUN echo "ServerAlias *" >> /etc/cups/cupsd.conf
+RUN echo "FileDevice yes" >> /etc/cups/cupsd.conf
+RUN sed -i 's/Order allow,deny/Order allow,deny\'$'\n  Allow all/g' /etc/cups/cupsd.conf
```
justinbastress: Is this missing an escape before the single quote after the $?

integration_tests/ipp/container/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+RUN sed -i 's/Listen localhost:631/Listen 0.0.0.0:631/g' /etc/cups/cupsd.conf
```
justinbastress: Using `sed` replacements ties this to a specific version of cups (and/or its packaging); this is fine for now, but maybe add a TODO comment to provide our own cupsd.conf rather than modifying the default.

integration_tests/ipp/container/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+RUN service cups restart
+
+# Try to make it act more container-y -- remove it from init.d and just run the daemon as the entrypoint
+#RUN service cupsd stop
```
justinbastress: I think the `service cupsd stop` probably needs to be un-commented -- if the `cupsd` daemon is running / listening on the configured port, that would conflict with the foreground process from the entrypoint.

But, since this is working, something else is clearly going on (maybe the service isn't being started in the first place?) -- so maybe for now just add a TODO comment.

 - [ ] Add TODO stop service / investigate why it works without stopping the service

integration_tests/ipp/container/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+#RUN update-rc.d -f cupsd remove
```
justinbastress: Same as above; the intent here is to not only stop the service, but prevent it from being started automatically.

integration_tests/ipp/container/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+RUN apt-get update
+RUN apt-get install -y cups
+RUN apt-get install -y cups-pdf
+RUN echo "ServerAlias *" >> /etc/cups/cupsd.conf
+RUN echo "FileDevice yes" >> /etc/cups/cupsd.conf
+RUN sed -i 's/Order allow,deny/Order allow,deny\'$'\n  Allow all/g' /etc/cups/cupsd.conf
+RUN sed -i 's/Listen localhost:631/Listen 0.0.0.0:631/g' /etc/cups/cupsd.conf
+RUN service cups restart
+
+# Try to make it act more container-y -- remove it from init.d and just run the daemon as the entrypoint
```
justinbastress: - [ ] Remove this comment

integration_tests/ipp/container/Dockerfile
flav: **Required** ([_pull request commenting with a purpose_](https://gist.github.com/andrewsardone/3fb9746e415d59784e97))

To keep things tidy (fewer disk layers), and following [best practices](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#minimize-the-number-of-layers) some of these commands should be doubled up into a single layer/`RUN` command. e.g.
```
RUN apt-get update \
    && apt-get install -y cups cups-pdf \
    …
```

modules/ipp/ipp.go
```diff
@@ -0,0 +1,71 @@
...
+	return b
+}
+
+//FIXME: Clean this up, and use binary package to handle endianness in the correct order
+//FIXME: Make sure uint isn't messing anything up here, since they should be signed.
+func attributeByteString(syntaxTag byte, name string, value string) []byte {
+	//special bytestring denoting value syntax
+	b := []byte{syntaxTag}
+	l := make([]byte, 2)
+	binary.PutUvarint(l, uint64(len(name)))
```
justinbastress: Do they actually use golang's varint format (high bit -> more bytes remaining), or is this just assuming len(name) < 127`? In the latter case, it is probably more appropriate to just put a fixed 8-bit size.

 - [ ] Add link to RFC section describing the encoding

modules/ipp/ipp.go
```diff
@@ -0,0 +1,71 @@
...
+	b = append(b, l...)
+	b = append(b, []byte(value)...)
+	return b
+}
+
+//TODO: Dynamically create nothing except uri?
+//Construct a minimal request that an IPP server will respond to
+func getPrinterAttributesRequest(uri string) bytes.Buffer {
+	var b bytes.Buffer
+	//version 2.1 (newest as of 2018)
```
justinbastress: - [ ] Link RFC section describing encoding

modules/ipp/ipp.go
```diff
@@ -0,0 +1,71 @@
...
+	b.Write([]byte{2, 1})
+	//operation-id = get-printer-attributes
+	b.Write([]byte{0, 0xb})
+	//request-id = 1
+	b.Write([]byte{0, 0, 0, 1})
+	//operation-attributes-tag = 1
+	b.Write([]byte{1})
+
+	//attributes-charset
```
justinbastress: What are the syntax tags here? Comments like you have above would be good, if they are going to be used in more than one place, you might create constants, maybe even an alias "type":

```
const SyntaxTagAttributesCharset uint8 = 0x47
```

or

```
type SyntaxTag uint8
const SyntaxTagAttributesCharset = SyntaxTag(0x47)
```

modules/ipp/scanner.go
```diff
@@ -0,0 +1,206 @@
...
+	"strings"
+	//"net"
+	//"net/url"
+	//"time"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/zmap/zgrab2"
+)
+
+const (
```
justinbastress: Go constants are typically of the form `ContentType` rather than `CONTENT_TYPE`; the golint will warn on the latter.

If the constants are copied from another API or an RFC, then keeping the well-known names is more important than language style, but for internal constants / constants that we're defining, it's probably best to use the go style.

 - [ ] `CONTENT_TYPE` -> `ContentType`

modules/ipp/scanner.go
```diff
@@ -0,0 +1,206 @@
...
+	CONTENT_TYPE string = "application/ipp"
+)
+
+//TODO: Tag relevant results and exlain in comments
+// ScanResults instances are returned by the module's Scan function.
+type ScanResults struct {
+	//TODO: Include a full response or at least a blob in the data
+	//Response *http.Response `json:"response,omitempty"`
```
justinbastress: This would be nice, at least for verbose mode. Including the requests could be interesting, too.

integration_tests/ipp/container/Dockerfile
clayzim: Will do. Thanks, flav.

modules/ipp/scanner.go
```diff
@@ -0,0 +1,206 @@
...
+
+	MajorVersion int8 `json:"version_major"`
+	MinorVersion int8 `json:"version_minor"`
+
+	VersionString string `json:"version_string,omitempty"`
+	CUPSVersion string `json:"cups_version,omitempty"`
+
+	//TODO: Uncomment this when implementing the TLS version of things
```
justinbastress: You could un-comment it now, since the `omitempty` will cause it to be omitted unless you explicitly set it.

integration_tests/ipp/container/Dockerfile
clayzim: @justinbastress I'm not 100% certain, but I don't believe so. It works as is, though looking at it now it definitely seems like it should have one.

modules/ipp/scanner.go
```diff
@@ -0,0 +1,206 @@
...
+	if https {
+		proto = "https"
+	} else {
+		proto = "http"
+	}
+	return proto + "://" + host + ":" + strconv.FormatUint(uint64(port), 10) + endpoint
+}
+
+func ippInContentType(resp http.Response) bool {
+	for _, t := range resp.Header["Content-Type"] {
```
justinbastress: Looking at https://github.com/zmap/zgrab2/blob/master/lib/http/header.go, that *might* work, but it looks like  `resp.Header.Get("Content-Type")` might be safer (though both are used in our code).

modules/ipp/scanner.go
```diff
@@ -0,0 +1,206 @@
...
+		if strings.Contains(t, CONTENT_TYPE) {
```
justinbastress: This is really nit-picky, but this would match e.g. `"application/ippimississ"`; there should be a library for checking content types, but if not, something like `t == CONTENT_TYPE or strings.StartsWith(t, CONTENT_TYPE + ";")` with a reference to the RFC would do in a pinch.

modules/ipp/scanner.go
```diff
@@ -0,0 +1,206 @@
...
+	host := target.Domain
+	if host == "" {
+		host = target.IP.String()
+	}
+	//FIXME: ?Should just use endpoint "/", since we get the same response as "/ipp" on CUPS??
+	uri := getIPPURL(scanner.config.IPPSecure, host, uint16(scanner.config.BaseFlags.Port), "/ipp")
+	b := getPrinterAttributesRequest(uri)
+	resp, err := http.Post(uri, CONTENT_TYPE, &b)
+	if err != nil {
+		//TODO: Create a descriptive error
```
justinbastress: You can use `zgrab2.TryGetScanStatus(err)` to auto-detect certain statuses, like timeouts / connection refused. This would be a good place to use that.

 - [ ] Detect the status

modules/ipp/scanner.go
```diff
@@ -0,0 +1,206 @@
...
+		return 0, 0, zgrab2.NewScanError(zgrab2.SCAN_UNKNOWN_ERROR, err)
+	}
+	if resp != nil && resp.Body != nil {
+		defer resp.Body.Close()
+	} else {
+		//TODO: Determine whether we need this error to avoid reading from Body
+		return 0, 0, zgrab2.NewScanError(zgrab2.SCAN_UNKNOWN_ERROR, err)
```
justinbastress: `err` is nil here; if there is no body, you can argue that's a `zgrab2.SCAN_PROTOCOL_ERROR`, assuming that an empty body is not allowed in IPP.

If empty bodies are allowed but indicate an error, `zgrab2.SCAN_APPLICATION_ERROR` might be more appropriate.

 - [ ] Remove `err`, confirm status

modules/ipp/scanner.go
```diff
@@ -0,0 +1,206 @@
...
+	//Check to make sure that the repsonse received is actually IPP
+	//Content-Type header matches is sufficient
+	//HTTP on port 631 is sufficient
+	//Still record data in the case of protocol error to see what that data looks like
+
+	//TODO: Record server-header version numbers
+	//protocols := resp.Header["Server"])
+	var version int16
+	if err := binary.Read(resp.Body, binary.BigEndian, &version); err != nil {
+		return 0, 0, zgrab2.NewScanError(zgrab2.SCAN_UNKNOWN_ERROR, err)
```
justinbastress: Same comment as above; if body is sent but smaller than 2 bytes, that is probably a `SCAN_PROTOCOL_ERROR`, maybe `SCAN_APPLICATION_ERROR`.

modules/ipp/scanner.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+	return int8(version >> 8), int8(version & 0xff), nil
```
justinbastress: Should these be `var version uint16` / `uint8(version >> 8)` etc? Or are negative versions allowed?

modules/ipp/scanner.go
```diff
@@ -0,0 +1,206 @@
...
+// Scan TODO: describe how scan operates
+//1. Send a request (currently get-printer-attributes)
+//2. Take in that response & read out version numbers
+func (scanner *Scanner) Scan(target zgrab2.ScanTarget) (zgrab2.ScanStatus, interface{}, error) {
+	// TODO: use Connection again, at least when implementing TLS
+	major, minor, err := scanner.grab(target)
+	//FIXME: Triggering even though error IS nil
+	//FIXME: This is a sloppy bodge to handle the issue
+	if major == 0 && minor == 0 && err != nil {
+		//TODO: Consider mimicking HTTP Scan's retryHTTPS functionality
```
justinbastress: If you want `grab()` to be able to return partial results, it should probably be returning a `*ScanResults` -- so you can check `if results == nil && err != nil` instead of e.g. `if major == 0 && minor == 0 && someString == "" && otherInt == 0 && err != nil`

modules/ipp/scanner.go
```diff
@@ -0,0 +1,206 @@
...
+		//TODO: Create more detailed error message?
+		return zgrab2.TryGetScanStatus(err), nil, err
+	}
+	results := &ScanResults{}
+	results.MajorVersion = major
+	results.MinorVersion = minor
+	return zgrab2.SCAN_SUCCESS, results, nil
```
justinbastress: It looks like you're throwing away `err` if it was able to detect a major/minor; it looks like the intent was something like `return zgrab2.TryGetScanStatus(err), results, err`; if `err == nil`, that will be identical to what you have now, but if there is an error, it will still return the partial results, alongside the error and a status.

zgrab2_schemas/zgrab2/ipp.py
```diff
@@ -0,0 +1,22 @@
...
+import zcrypto_schemas.zcrypto as zcrypto
+import zgrab2
+
+ipp_scan_response = SubRecord({
+    "result": SubRecord({
+        "version_major": Signed8BitInteger(),
+        "version_minor": Signed8BitInteger(),
+        "version_string": String(),
+        "cups_version": String(),
+        #"tls": zgrab2.tls_log,
```
justinbastress: Since this isn't required, you could un-comment it and everything would still work.

modules/ipp/scanner.go
clayzim: I don't believe that negative values are ever used, but the RFC specifies that _"Every integer in this encoding MUST be encoded as a signed integer using two's-complement binary encoding with big-endian format"_ (Source: https://tools.ietf.org/html/rfc8010#section-3)

justinbastress: Got it -- I'd say just drop that link / sentence into a comment.

Also, I don't recall off-hand how `>>` works with signed integers (whether it keeps / extends the sign), but to be safe / explicit, it should probably be two `Read`s (of `int8`s), or change `version` to a `uint16` so that there is no ambiguity (since version is just being treated as a two-byte buffer anyway).

modules/ipp/ipp.go
```diff
@@ -0,0 +1,76 @@
...
+//func ReadResponse(body *io.ReadCloser) *ScanResults {
+//	result := &ScanResults{}
+//
+//}
+
+// Returns a byte-encoded "attribute-with-one-value" with the provided "value-tag", "name", and "value"
+// attribute-with-one-value encoding described at https://tools.ietf.org/html/rfc8010#section-3.1.4
+// Example (runnable from ipp_test.go):
+//   Input: 0x47, "attributes-charset", "us-ascii"
+//   Output: [71 0 18 97 116 116 114 105 98 117 116 101 115 45 99 104 97 114 115 101 116 0 8 117 115 45 97 115 99 105 105]
```
justinbastress: For some reason, golang decides to dump bytes in decimal -- the output of `hex.Dump()` is usually much more helpful.

modules/ipp/ipp.go
```diff
@@ -0,0 +1,76 @@
...
+// TODO: Should return an error when fed an invalid valueTag?
+// TODO: Determine whether this should remain public. Currently is for Testable Example
```
justinbastress: I know for normal unit tests, if your test code is in the same package, functions don't need to be public to be tested; is that not the case for examples?

modules/ipp/ipp.go
```diff
@@ -0,0 +1,76 @@
...
+func AttributeByteString(valueTag byte, name string, value string) []byte {
+	//special byte denoting value syntax
+	b := []byte{valueTag}
+
+	//append 16-bit signed int denoting name length
+	l := new(bytes.Buffer)
```
justinbastress: Even though the arguments aren't coming from external sources, it still might be a good idea to check that `len(name)`/`len(value)` will fit into an `int16` (so < 0x8000)

modules/ipp/ipp.go
```diff
@@ -0,0 +1,76 @@
...
+// Returns a byte-encoded "attribute-with-one-value" with the provided "value-tag", "name", and "value"
+// attribute-with-one-value encoding described at https://tools.ietf.org/html/rfc8010#section-3.1.4
+// Example (runnable from ipp_test.go):
+//   Input: 0x47, "attributes-charset", "us-ascii"
+//   Output: [71 0 18 97 116 116 114 105 98 117 116 101 115 45 99 104 97 114 115 101 116 0 8 117 115 45 97 115 99 105 105]
+// TODO: Should return an error when fed an invalid valueTag?
+// TODO: Determine whether this should remain public. Currently is for Testable Example
+func AttributeByteString(valueTag byte, name string, value string) []byte {
+	//special byte denoting value syntax
+	b := []byte{valueTag}
```
justinbastress: It looks like this could be simplified by making a single `bytes.Buffer` and writing to that everywhere, instead of going back and forth `append`ing to `b`

modules/ipp/ipp.go
```diff
@@ -0,0 +1,76 @@
...
+	return b
+}
+
+
+// IPP request encoding described at https://tools.ietf.org/html/rfc8010#section-3.1.1
+//TODO: Store everything except uri statically?
+//Construct a minimal request that an IPP server will respond to
+func getPrinterAttributesRequest(uri string) bytes.Buffer {
+	var b bytes.Buffer
+	//version 2.1 (newest as of 2018)
```
justinbastress: Maybe include a sentence describing why we choose the newest version -- e.g. how do older versions respond to a too-new version? How do newer versions respond to an older version?

integration_tests/ipp/container/Dockerfile
```diff
@@ -1,16 +1,17 @@
...
-RUN apt-get install -y cups
-RUN apt-get install -y cups-pdf
-RUN echo "ServerAlias *" >> /etc/cups/cupsd.conf
-RUN echo "FileDevice yes" >> /etc/cups/cupsd.conf
-RUN sed -i 's/Order allow,deny/Order allow,deny\'$'\n  Allow all/g' /etc/cups/cupsd.conf
-RUN sed -i 's/Listen localhost:631/Listen 0.0.0.0:631/g' /etc/cups/cupsd.conf
+RUN apt-get update && apt-get install -y \
+  cups \
+  cups-pdf
+# TODO: Provide a pre-built cupsd.conf rather than relying on modifying the default config file
```
justinbastress: It looks like this is done

clayzim: Oh yeah, that is done. Oops.

justinbastress: By the way, you can merge whenever you're ready -- none of the last round of comments were blocking.

PR #137, opened 2018-06-20
Feature/create ipp module

clayzim:
Implements TLS and retrying as TLS if plaintext is failing.
Introduces a dependency on zgrab2/lib/http and copy-pastes some private utility functions from HTTP module.
Creates separate TLS and plaintext IPP containers, each with a file printer.
Attempts to make IPP requests with older versions if it's receiving version-not-supported error.

## How to Test
`make integration-test`

## Notes & Caveats
Future work:
Output schema will be modified (if appropriate) in response to 1% scan.
Less code would be duplicated if we move a few critical functions from the HTTP module into zgrab2/lib/http (issue #138)

## Issue Tracking
https://github.com/zmap/zgrab2/issues/138 - library HTTP code


integration_tests/ipp/container-cups-tls/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM zgrab2_service_base:latest
+
+RUN apt-get update && apt-get install -y \
+  cups \
+  cups-pdf
+WORKDIR /etc/cups
+COPY cupsssl.conf cupsd.conf
+# TODO: Provide a pre-built cups-pdf.conf
```
justinbastress: Can the TODO and commented RUN be removed?

integration_tests/ipp/container-cups-tls/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+#RUN service cups restart
+
+# TODO: Actually stop service; see why this works without stopping it
```
justinbastress: Looks like this is done?

integration_tests/ipp/container-cups-tls/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+RUN service cups stop
+# TODO: Actually prevent cups from being started automatically
```
justinbastress: Is cups still being started automatically despite the update-rc.d? There may be a better way to do this, but if that's the concern the TODO should probably be reworded

integration_tests/ipp/container-cups-tls/cupsssl.conf
```diff
@@ -0,0 +1,152 @@
...
+
+# Only listen for connections from the local machine.
+Listen /var/run/cups/cups.sock
+
+# Show shared printers on the local network.
+Browsing Off
+BrowseLocalProtocols dnssd
+
+# Default authentication type, when authentication is required...
+DefaultAuthType Basic
```
justinbastress: Future -- what other options are there for DefaultAuthType?

integration_tests/ipp/container-cups-tls/cupsssl.conf
```diff
@@ -0,0 +1,152 @@
...
+
+# Web interface setting...
+WebInterface Yes
```
justinbastress: If WebInterface is `No`, does that affect the scan?

integration_tests/ipp/container-cups/Dockerfile
```diff
@@ -3,16 +3,15 @@ FROM zgrab2_service_base:latest
 RUN apt-get update && apt-get install -y \
   cups \
   cups-pdf
-# TODO: Provide a pre-built cupsd.conf rather than relying on modifying the default config file
 WORKDIR /etc/cups
 COPY cupsd.conf cupsd.conf
 # TODO: Provide a pre-built cups-pdf.conf
-RUN service cups restart
+#RUN service cups restart
```
justinbastress: Same comments as with the other Dockerfile

integration_tests/ipp/container-cups-tls/cupsssl.conf
clayzim: It shouldn't affect anything. The scan is sending get-printer-attributes requests (and optionally CUPS-specific CUPS-get-printers requests), which don't rely on the web interface. It's mostly that way because I modified as few of the default settings as possible.
It's possible that we could scrape information off of the web interface if it's enabled from all addresses.

integration_tests/ipp/test.sh
```diff
@@ -5,32 +5,91 @@ MODULE_DIR=$(dirname $0)
...
+    echo "ipp/test: Tests runner for ipp_cups"
+
+    CONTAINER_NAME="zgrab_ipp_cups-tls" $ZGRAB_ROOT/docker-runner/docker-run.sh ipp --timeout 3 --ipps --verbose > "$OUTPUT_ROOT/cups-tls.json"
+    # FIXME: No good reason to use a tmp file & saved file, b/c I'm not testing any failure states yet
+    #CONTAINER_NAME="zgrab_ipp_cups-tls" $ZGRAB_ROOT/docker-runner/docker-run.sh ipp --timeout 3 --ipps --verbose > out.tmp
+    major=$($ZGRAB_ROOT/jp -u data.ipp.result.version_major < "$OUTPUT_ROOT/cups-tls.json")
+    minor=$($ZGRAB_ROOT/jp -u data.ipp.result.version_minor < "$OUTPUT_ROOT/cups-tls.json")
+    response=$($ZGRAB_ROOT/jp -u data.ipp.result.response < "$OUTPUT_ROOT/cups-tls.json")
+    cups=$($ZGRAB_ROOT/jp -u data.ipp.result.cups_version < "$OUTPUT_ROOT/cups-tls.json")
+    tls=$($ZGRAB_ROOT/jp -u data.ipp.result.tls < "$OUTPUT_ROOT/cups-tls.json")
```
justinbastress: Here it might be safer to check for a field in e.g. `data.mysql.result.tls.handshake_log.server_certificates.certificate.parsed.subject_dn` -- you still might have some of the `tls` structure (e.g. the `client_hello`) even if the server doesn't give a valid response.

integration_tests/ipp/test.sh
```diff
@@ -5,32 +5,91 @@ MODULE_DIR=$(dirname $0)
...
+    fi
+    if [ $tls = "null" ]; then
+        echo "ipp/test: No TLS handshake logged"
+    fi
+}
 
 echo "ipp/test: Testing IPP on $CONTAINER_NAME..."
 # TODO FIXME: Add any necessary flags or additional tests
-CONTAINER_NAME=$CONTAINER_NAME $ZGRAB_ROOT/docker-runner/docker-run.sh ipp > $OUTPUT_FILE
+#CONTAINER_NAME=$CONTAINER_NAME $ZGRAB_ROOT/docker-runner/docker-run.sh ipp > $OUTPUT_FILE
```
justinbastress: This has been permanently replaced by `test_cups(_tls)` right?

modules/ipp/ipp.go
```diff
@@ -3,74 +3,119 @@ package ipp
...
+func AttributeByteString(valueTag byte, name string, value string, target *bytes.Buffer) error {
 	//special byte denoting value syntax
-	b := []byte{valueTag}
+	binary.Write(target, binary.BigEndian, valueTag)
 
-	//append 16-bit signed int denoting name length
-	l := new(bytes.Buffer)
-	binary.Write(l, binary.BigEndian, int16(len(name)))
-	b = append(b, l.Bytes()...)
+	if len(name) < (1 << 16) {
```
justinbastress: Since it's signed, shouldn't it be ` < 0x8000` / `< (1 << 15)`?

justinbastress: Actually, it looks like go provides `math.MaxInt16` etc -- https://golang.org/pkg/math/#pkg-constants

modules/ipp/ipp.go
```diff
@@ -3,74 +3,119 @@ package ipp
...
+	return nil
 }
 
+func convertURIToIPP(uri string) string {
+	if strings.HasPrefix(uri, "http://") || strings.HasPrefix(uri, "https://") {
+		uri = strings.Replace(uri, "http", "ipp", 1)
+	}
+	// TODO: RFC claims that literal IP addresses are not valid uri's, but Wireshark IPP Capture example uses them
+	// (Source: https://wiki.wireshark.org/SampleCaptures?action=AttachFile&do=view&target=ipp.pcap)
+	sections := strings.Split(uri, "/")
```
justinbastress: Future -- look into `net.url.Parse()` / `URL.String()` for parsing / replacing / re-encoding URL fields?

modules/ipp/scanner.go
```diff
@@ -4,40 +4,73 @@ package ipp
...
 type ScanResults struct {
 	//TODO: ?Include the request sent as well??
-	//TODO: Include a full response or at least a blob in the data (at least in verbose mode)
-	//Response *http.Response `json:"response,omitempty"`
+	Response     *http.Response `json:"response,omitempty" zgrab:"debug"`
+	CUPSResponse *http.Response `json:"cups_response,omitempty" zgrab:"debug"`
 
-	MajorVersion int8 `json:"version_major"`
-	MinorVersion int8 `json:"version_minor"`
+	// RedirectResponseChain is non-empty if the scanner follows a redirect.
```
justinbastress: It looks like these new fields need to be added to https://github.com/zmap/zgrab2/blob/46029a8176d3a3d0e141fa889795d7c5678f3e37/zgrab2_schemas/zgrab2/ipp.py

modules/ipp/scanner.go
```diff
@@ -126,86 +172,394 @@ func (scanner *Scanner) GetPort() uint {
...
+	res.Body.Close()
+	res.Body = ioutil.NopCloser(b)
+	return b
+}
+
+// FIXME: add proper error handling to this
+// TODO: Support reading from multiple instances of the same attribute in a response
+func readAttributeFromBody(attrString string, body *[]byte) ([][]byte, error) {
+	attr := []byte(attrString)
+	interims := bytes.Split(*body, attr)
```
justinbastress: Is this valid? Like, what if there was an attribute `max-supported-cups-version` -- that would get picked up by `readAttributeFromBody` wouldn't it?

modules/ipp/scanner.go
```diff
@@ -126,86 +172,394 @@ func (scanner *Scanner) GetPort() uint {
...
+				continue
+			}
+			if err := binary.Read(buf, binary.BigEndian, &nameLength); err != nil {
+				//Couldn't read next nameLength
+				//break
+				// TODO: Conclude correct error behavior for all of these blocks
+				continue
+			}
+		}
+		return vals, nil // FIXME: This should be whatever error occurred during a break case
```
justinbastress: In that case, could you do `return vals, <the error>` instead of `break`, since there is no further handling?

modules/ipp/scanner.go
clayzim: You're definitely right. My first thought here is to write a bare bones IPP response parser. Does that seem appropriate, or is there an alternative plan of action you'd suggest?

modules/ipp/scanner.go
```diff
@@ -126,86 +172,394 @@ func (scanner *Scanner) GetPort() uint {
...
-	uri := getIPPURL(scanner.config.IPPSecure, host, uint16(scanner.config.BaseFlags.Port), "/ipp")
-	b := getPrinterAttributesRequest(uri)
+	//The attribute was not present
+	return nil, errors.New("Attribute \"" + attrString + "\" not present.")
+}
 
-	resp, err := http.Post(uri, ContentType, &b)
+func versionNotSupported(resp *http.Response, scanner *Scanner) bool {
+	if resp != nil && resp.Body != nil {
+		buf := bufferFromBody(resp, scanner)
```
justinbastress: Am I missing something, or won't this wipe out `resp.Body` -- so any subsequent calls (like `storeBody()` on 299) will get nothing back?

The way I read it, it seems like `storeBody()` should be called first, then `versionNotSupported` called on that -- but I may be missing something.

clayzim: I've been struggling with how to handle that all correctly. `bufferFromBody()` returns a buffer filled with the contents of the body, but should assign a new io.ReadCloser with the original contents to `resp.Body`*. My understanding was that subsequent calls reading from `resp.Body` (like `storeBody()`) should find a non-empty ReadCloser and be fine. But I've been having some issues with this. Basically `versionNotSupported` was exhibiting incorrect behavior when executed after `storeBody()`. So something's definitely off.

*(Technically it's only the first `maxReadLen` KB of `resp.Body` because that's all we'll read anyway).

modules/ipp/scanner.go
```diff
@@ -126,86 +172,394 @@ func (scanner *Scanner) GetPort() uint {
...
+		defer cupsResp.Body.Close()
+	}
+	// Store data into BodyText and BodySHA256 of cupsResp
+	storeBody(cupsResp, scanner)
+
+	bodyBytes := []byte(cupsResp.BodyText)
+	cupsVersions, _ := readAttributeFromBody(CupsVersion, &bodyBytes)
+	if len(cupsVersions) > 0 {
+		// Report this cupsVersion b/c it's more detailed than the one found in Server header
+		scan.results.CUPSVersion = strings.Split(scan.results.CUPSVersion, "/")[0] + "/" + string(cupsVersions[0])
```
justinbastress: Should these just be different response fields? Maybe `CUPSVersion` is the one from the header, the attribute version could be included alongside other attributes?

modules/ipp/scanner.go
```diff
@@ -126,86 +172,394 @@ func (scanner *Scanner) GetPort() uint {
...
+	// Send get-printer-attributes request to the host, preferably a print server
+	body := getPrinterAttributesRequest(version.Major, version.Minor, scan.url)
+	request, err := http.NewRequest("POST", scan.url, body)
+	if err != nil {
+		return zgrab2.NewScanError(zgrab2.SCAN_UNKNOWN_ERROR, err)
+	}
+	request.Header.Set("Accept", "*/*")
+	request.Header.Set("Content-Type", ContentType)
+	resp, err := scan.client.Do(request)
+	if versionNotSupported(resp, scanner) {
```
justinbastress: Same comment as above re: versionNotSupported reading / replacing the body

modules/ipp/scanner.go
```diff
@@ -126,86 +172,394 @@ func (scanner *Scanner) GetPort() uint {
...
+		return zgrab2.NewScanError(zgrab2.SCAN_APPLICATION_ERROR, ErrVersionNotSupported)
+	}
+	//Store response regardless of error in request, because we may have gotten something back
+	scan.results.Response = resp
 	if resp != nil && resp.Body != nil {
 		defer resp.Body.Close()
```
justinbastress: This should probably go right after you get `resp` back -- in e.g. line 349 it could be leaked (if it weren't already reading / closing it)

modules/ipp/scanner.go
```diff
@@ -126,86 +172,394 @@ func (scanner *Scanner) GetPort() uint {
...
+		if strings.HasPrefix(strings.ToUpper(p), "IPP/") {
+			scan.results.VersionString = p
+			protocol := strings.Split(p, "/")[1]
+			components := strings.Split(protocol, ".")
+			// Reads in signed integers because "every integer MUST be encoded as a signed integer"
+			// (Source: https://tools.ietf.org/html/rfc8010#section-3)
+			var major, minor int8
+			if len(components) >= 1 {
+				if val, err := strconv.Atoi(components[0]); err != nil {
+					// TODO: Handle error
```
justinbastress: I'd say for now, add a debug-level log, with a TODO to remove it -- that could help debug any issues we in the 1% scan.

That would go for any other places where there are potentially errors that you can't do anything about, but indicate unexpected behavior.

modules/ipp/scanner.go
```diff
@@ -126,86 +172,394 @@ func (scanner *Scanner) GetPort() uint {
...
-	results.MinorVersion = minor
-	// FIXME: Triggering even though error IS nil
-	// FIXME: This is a sloppy bodge to handle the issue
-	if major == 0 && minor == 0 && err != nil {
-		// TODO: Consider mimicking HTTP Scan's retryHTTPS functionality
-		return zgrab2.TryGetScanStatus(err), results, err
-	}
-	return zgrab2.SCAN_SUCCESS, results, nil
+	scan := scanner.newIPPScan(&target)
+	//defer scan.Cleanup()
```
justinbastress: Add a TODO to re-add this?

integration_tests/ipp/container-cups-tls/cupsssl.conf
justinbastress: If the default is on, then it's not a big deal.

A future TODO might be detecting if the web interface is enabled; that could be an interesting metric ("how many publicly-exposed print servers also public expose the web management interface?"), but probably very interesting to implement.

modules/ipp/scanner.go
justinbastress: Could you call `storeBody()` once at the start, then always use `response.BodyText`? I may be missing some problem with that, but that was what I was expecting to see.

clayzim:
Pushing some corrections. I'm gonna figure out logging unexpected errors in an additional results field tomorrow.

modules/ipp/ipp.go
```diff
@@ -18,14 +19,15 @@ func AttributeByteString(valueTag byte, name string, value string, target *bytes
 	//special byte denoting value syntax
 	binary.Write(target, binary.BigEndian, valueTag)
 
-	if len(name) < (1 << 16) {
+	if len(name) <= math.MaxInt16 && len(name) >= math.MinInt16  {
```
justinbastress: This will do it, but if `len(name) < 0` we have other issues

modules/ipp/scanner.go
```diff
@@ -227,47 +228,43 @@ func readAttributeFromBody(attrString string, body *[]byte) ([][]byte, error) {
...
+		return vals, nil
 	}
 	//The attribute was not present
 	return nil, errors.New("Attribute \"" + attrString + "\" not present.")
 }
 
-func versionNotSupported(resp *http.Response, scanner *Scanner) bool {
-	if resp != nil && resp.Body != nil {
-		buf := bufferFromBody(resp, scanner)
+func versionNotSupported(body *string, scanner *Scanner) bool {
```
justinbastress: FYI, passing a string "by value" doesn't mean that it's going to copy the entire body -- it's sort of like Java, where it copies the length / pointer.

integration_tests/ipp/test.sh
```diff
@@ -29,8 +29,9 @@ function test_cups() {
         echo "ipp/test: Incorrect minor version. Expected 1, got $minor"
         exit 1
     fi
-    if ! [ $cups = "CUPS/2.1.3" ]; then
+    if ! [ $cups = "CUPS/2.1" ]; then
         echo "ipp/test: Incorrect CUPS version. Expected CUPS/2.1.3, got $cups"
```
justinbastress: Condition / error message mismatch

integration_tests/ipp/test.sh
```diff
@@ -55,11 +56,13 @@ function test_cups_tls() {
...
```
justinbastress: Condition/error message mismatch

modules/ipp/scanner.go
```diff
@@ -289,45 +297,39 @@ func (scanner *Scanner) augmentWithCUPSData(scan *scan, target *zgrab2.ScanTarge
...
-	if len(uris) > 0 {
-		scan.results.PrinterURI = string(uris[0])
+	// Write reported printer URI to results object
+	if uris, err := readAttributeFromBody(PrinterURISupported, &bodyBytes); err != nil {
+		scan.log.WithFields(log.Fields{
+			"error": err,
+			"attribute": PrinterURISupported,
+		}).Warn("Failed to read attribute.")
+	} else if len(uris) > 0 {
+		scan.results.AttributePrinterURI = string(uris[0])
```
justinbastress: Future -- is it permissible for multiple URIs to be returned? If so, it may be helpful to return them all (on the other hand, if it's just an artifact of the API always returning an array, ignore this)

modules/ipp/scanner.go
```diff
@@ -409,8 +417,12 @@ func (scanner *Scanner) Grab(scan *scan, target *zgrab2.ScanTarget, version *ver
 		}
 		if strings.HasPrefix(strings.ToUpper(p), "CUPS/") {
 			scan.results.CUPSVersion = p
-			/*err :=*/ scanner.augmentWithCUPSData(scan, target, version)
-			// TODO: Log error
+			err := scanner.augmentWithCUPSData(scan, target, version)
+			if err != nil {
+				scan.log.WithFields(log.Fields{
+					"error": err,
+				}).Warn("Failed to augment with CUPS-get-printers request.")
```
justinbastress: Since these scans get run over the entire public IPv4 space, we typically would only want info+-level logs to appear when there is a problem with the scan itself, rather than a problem with an individual host.

The exception would be if debugging / verbose logging is enabled. So, I see two easy options -- move all of the `*.Warn()` etc to `*.Debug()`, or, set the log level of your logger in such a way that it only logs if `Flags.Verbose` is set (I think `scan.log.SetLevel(logrus.ErrorLevel)` should work, but I'm not certain).

PR #140, opened 2018-06-26
(IPP) improve scan output

clayzim:
Correctly determines whether to include ScanResults object in output json.
Rejects successful responses which are identifiable as non-IPP.
Reduces code duplication by refactoring several re-usable segments into independent functions.
Greatly simplifies IPP zgrab2 schema by depending upon HTTP zgrab2 schema.

## How to Test

`make integration-test`

## Notes & Caveats

Future Work:
Add support for Upgrade to HTTPS.

## Issue Tracking

N/A

### DIFF
--- a/modules/ipp/scanner.go
+++ b/modules/ipp/scanner.go
@@ -439,7 +439,7 @@ func (scanner *Scanner) Grab(scan *scan, target *zgrab2.ScanTarget, version *ver
 	// TODO: Cite RFC justification for this
 	// Reject successful responses which specify non-IPP MIME mediatype (ie: text/html)
 	if isIPP, _ := hasContentType(resp, ContentType);
-		resp.StatusCode == 200 && resp.Header.Get("Content-Type") != "" && !isIPP {
+	   resp.StatusCode == 200 && resp.Header.Get("Content-Type") != "" && !isIPP {
 		// TODO: Log error if any
 		return zgrab2.NewScanError(zgrab2.SCAN_PROTOCOL_ERROR, errors.New("application/ipp not present in Content-Type header."))
 	}
--- a/modules/ipp/scanner.go
+++ b/modules/ipp/scanner.go
@@ -42,7 +42,7 @@ var (
 	// TODO: Explain this error
 	ErrVersionNotSupported = errors.New("IPP version not supported")
 
-	Versions = []version {{Major: 2, Minor: 1}, {Major: 2, Minor: 0}, {Major: 1, Minor: 1}, {Major: 1, Minor: 0},}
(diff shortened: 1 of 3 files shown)


modules/ipp/scanner.go
```diff
@@ -566,49 +551,60 @@ func (scanner *Scanner) newIPPScan(target *zgrab2.ScanTarget) *scan {
...
 
+// TODO: Do you want to retry with TLS for all versions? Just one's you've already tried? Haven't tried? Just the same version?
+func (scanner *Scanner) tryGrabForVersions(target *zgrab2.ScanTarget, versions *[]version) (*scan, *zgrab2.ScanError) {
+	scan := scanner.newIPPScan(target)
+	// TODO: Implement scan.Cleanup()
+	var err *zgrab2.ScanError
+	for i := 0; i < len(*versions); i++ {
+		err = scanner.Grab(scan, target, &(*versions)[i])
+		if err != nil && err.Err == ErrVersionNotSupported && i < len(*versions)-1 {
+			continue
```
justinbastress: This construction is a little weird -- I'm guessing the reasoning for `if (a) { continue } break` rather than `if !a { break }` is `!a` is kind of clumsy?

modules/ipp/scanner.go
```diff
@@ -566,49 +551,60 @@ func (scanner *Scanner) newIPPScan(target *zgrab2.ScanTarget) *scan {
...
-		}
-	}
+	// Try all known IPP versions from newest to oldest until we reach a supported version
+	scan, err := scanner.tryGrabForVersions(&target, &Versions)
 	if err != nil {
-		// Adapted from http module's RetryHTTPS logic
+		// If versionNotSupported error was confirmed, the scanner was connecting w/o TLS, so don't retry
+		// Same goes for a protocol error of any kind. It means we got something back but it didn't conform.
+		if err.Status == zgrab2.SCAN_APPLICATION_ERROR || err.Status == zgrab2.SCAN_PROTOCOL_ERROR {
+			return err.Unpack(&scan.results)
```
justinbastress: It looks like this will always return a non-nil result (so always indicating a detection in these cases) -- is that the intent? Since `scan.results` is (IIRC) a `ScanResults` and not a `*ScanResults`, its default/not-yet-set value will be `{}`, not `nil`.

zakird:
@clayzim can you confirm justin's question? Trying to get things merged.

modules/ipp/scanner.go
clayzim: Oops sorry for not responding. My intention is to return the partial result despite the error in this case, because each of these statuses indicate a positive identification as IPP but with an error.

zakird:
Thank you @clayzim !

PR #141, opened 2018-06-27
Add support for Tags and Triggers

jhalderm:
This PR adds support for a new pair of concepts, tags and triggers, that allow a ZGrab instance to more flexibly scan multiple protocols.

A tag is a string associated with a grab target.  Any input line can end in a tag, which is written preceded by an equals sign, as follows:

```
141.212.113.199 =tagA
216.239.38.21, censys.io =tagB
```

A trigger is a new parameter associated with each scan module.  When `--trigger=X` is specified, the scan module will only be invoked on targets with tag `X`.  Likewise, inputs containing a tag will not be processed by scan modules that do not include a corresponding trigger.

Tags and triggers are particularly powerful when combined with ZGrab's multiple command feature.  For instance, invoking `zgrab2 multiple` with the following configuration will perform an SSH grab on the first target above and an HTTP grab on the second target:

```
[ssh]
trigger="tagA"
name="ssh21"
port=22

[http]
trigger="tagB"
name="http80"
port=80
```

## Notes & Caveats

ZGrab2's multiple command support is currently hobbled by https://github.com/zmap/zflags/issues/3.  This may cause problems during testing.

### DIFF
--- a/processing.go
+++ b/processing.go
@@ -25,17 +25,25 @@ type Grab struct {
 type ScanTarget struct {
 	IP     net.IP
 	Domain string
+	Tag    string
 }
 
 func (target ScanTarget) String() string {
 	if target.IP == nil && target.Domain == "" {
 		return "<empty target>"
-	} else if target.IP != nil && target.Domain != "" {
-		return target.Domain + "(" + target.IP.String() + ")"
-	} else if target.IP != nil {
-		return target.IP.String()
 	}
-	return target.Domain
+	res := ""
+	if target.IP != nil && target.Domain != "" {
+		res = target.Domain + "(" + target.IP.String() + ")"
+	} else if target.IP != nil {
+		res = target.IP.String()
+	} else {
+		res = target.Domain
+	}
+	if target.Tag != "" {
+		res += " =" + target.Tag
+	}
+	return res
 }
 
 // Open connects to the ScanTarget using the configured flags, and returns a net.Conn that uses the configured timeouts for Read/Write operations.
@@ -81,6 +89,11 @@ func grabTarget(input ScanTarget, m *Monitor) []byte {
 	moduleResult := make(map[string]ScanResponse)
 
 	for _, scannerName := range orderedScanners {
+		scanner := scanners[scannerName]
+		trigger := (*scanner).GetTrigger()
+		if (input.Tag != "" || trigger != "") && input.Tag != trigger {
+			continue
+		}
 		defer func(name string) {
 			if e := recover(); e != nil {
 				log.Errorf("Panic on scanner %s when scanning target %s: %#v", scannerName, input.String(), e)
@@ -88,7 +101,6 @@ func grabTarget(input ScanTarget, m *Monitor) []byte {
 				panic(e)
 			}
 		}(scannerName)
-		scanner := scanners[scannerName]
 		name, res := RunScanner(*scanner, m, input)
 		moduleResult[name] = res
 		if res.Error != nil && !config.Multiple.ContinueOnError {
@@ -182,7 +194,7 @@ func Process(mon *Monitor) {
 			log.Error(err)
 		}
 		st := strings.TrimSpace(string(obj))
-		ipnet, domain, err := ParseTarget(st)
+		ipnet, domain, tag, err := ParseTarget(st)
 		if err != nil {
 			log.Error(err)
 			continue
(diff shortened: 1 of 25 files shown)


zakird:
Overall this looks good to me. The feature makes sense.

I'm not a fan of introducing a custom file syntax for ZGrab. The rest of the toolchain is standardized on CSV and JSON, which can be using produced/consumed with standard language libraries. Can we switch to either use CSV (with a third column that contains a list of tags/triggers) or JSON?

processing.go
```diff
@@ -81,14 +89,18 @@ func grabTarget(input ScanTarget, m *Monitor) []byte {
 	moduleResult := make(map[string]ScanResponse)
 
 	for _, scannerName := range orderedScanners {
+		scanner := scanners[scannerName]
+		trigger := (*scanner).GetTrigger()
+		if (input.Tag != "" || trigger != "") && input.Tag != trigger {
```
justinbastress: When looking over this last night, I read this as `if (input.Tag != "" && trigger != "") && input.Tag != trigger`, which seems more like the behavior I would expect -- namely, no-tag-matches-everything (bolstered by the observation that as it stands, the first clause is moot -- if it is false, i.e. both are empty, then the second clause is false).

This change would allow you to reuse tagged input with un-tagged scans, and probably more importantly, tagged scans with un-tagged input.

It would mean that if you re-use untagged input with a tagged scanner, all modules (rather than just untagged modules) would hit every host -- but in that case, suffixing each input line with a dummy tag would be a trivial workaround to get this behavior back.

jhalderm: Thanks for the thoughtful review, Justin.

I considered no-tag-matches-everything, but that seems unsafe as a default behavior.  Suppose someone accidentally omits the trigger when adding a new scanner to a configuration with dozens of tags.  They'll make boatloads of spurious connections, and might not notice unless they're paying attention to the failure rate.  With strict-tag-equals-trigger, the error is obvious immediately, since the scan never triggers.

What do others think?

jhalderm: Amended PR to fix the moot clause in the conditional.  Good catch Justin!

justinbastress: That makes sense -- basically, if you are using triggers, it is preferable for a misconfigured scanner to scan no hosts than for it to scan hosts that it was not intended to scan.

jhalderm:
@zakird, good point about the input format.  

Let's try to stay close to the existing CSV-like format.  Currently ZGrab accepts these cases:

```
IP
DOMAIN
IP, DOMAIN
CIDR
```

Unfortunately, we can't just add `[, TAG]` to each case, since we need to distinguish `IP, DOMAIN` from `IP, TAG`.

One option is to require the tag to always be column 3, e.g. `IP,,TAG`.  That would work, but it's inconsistent with how we handle `DOMAIN`, which can be column 1 or column 2.

Alternatively, we could define three fixed columns:  `IP/CIDR, DOMAIN, TAG`.  To avoid breaking compatibility, we can also accept lines that contain only DOMAIN, as a special case.

What do you think?

zakird:
I like the option of requiring tag to be column three. If multiple columns passed in, they must follow the defined standard, and then still allowing the backwards compatibility edge case where you can pass in anything as a single column, whether that be IP, domain, CIDR.

jhalderm:
Great.  I'll implement that as part of this PR.

jhalderm:
CSV input is done.  @zakird, have a look?

zakird:
@jhalderm fyi tests are failing on this

zakird:
Note: we should update readme for this

justinbastress:
Going through integration test output, the initial error is `dial tcp: lookup \u003cnil\u003e: no such host"` -- so it's apparently either getting either nil or the string `"<nil>"` for a host/IP in the `ScanTarget`.

jhalderm:
I'll look into the failing tests. Strange that it's only mysql.

justinbastress:
* Confirmed that it's not just MySQL, it's just MySQL is the only one where it's dumping the output
 * There do seem to be two scanners that are NOT affected: HTTP and IPP (which uses HTTP)
 * Adding a comma (which shouldn't be needed anyway) doesn't fix it
 * Everything does work if you just use an IP address instead of / in addition to a hostname

The fact that HTTP is unaffected seems like it must be related to the fact that it's the only scanner that actually uses the hostname (or, it is affected and it's just not reporting the error correctly)

justinbastress:
OK, duh -- `ScanTarget.Open()` was only using `target.IP`, assuming it was always filled in (even if it was actually filled in with a domain). Fixed in #142

jhalderm:
Justin, good catch!  I don't understand why it was working before though.

justinbastress:
Now it really is fixed except for one scanner -- because SSH was not using `ScanTarget.Open`, but rather dialing directly with `IP.String()`. Pushing a fix momentarily.

justinbastress:
(Future work) While we're on the topic of updating the input format -- @zakird, didn't you mention adding optional explicit ports for individual hosts to the input?

jhalderm:
Nice, thanks!  I can update the readme and merge this tomorrow.

PR #143, opened 2018-06-29
Feature/IPP: Fix Retry-TLS and Collect All IPP Attributes

clayzim:
Fixes retry-TLS. The retry occurs if and only if the plaintext connection fails.

Collects all attributes returned in any IPP responses received.

## How to Test

`make integration-test`

## Notes & Caveats

Future Work:
Add support for Upgrade to HTTPS.
Add support for proper redirect handling. (Currently reporting application error for non-200 status code.)

## Issue Tracking

N/A

### DIFF
--- a/modules/ipp/scanner.go
+++ b/modules/ipp/scanner.go
@@ -281,28 +281,35 @@ func readAllAttributes(body []byte) ([]*Attribute) {
 	return attrs
 }
 
-func (scan *scan) tryReadAttributes(body string) {
-	bodyBytes := []byte(body)
-	if bytes.Contains(bodyBytes, []byte(AttributesCharset)) {
-		scan.results.Attributes = append(scan.results.Attributes, readAllAttributes(bodyBytes)...)
+func (scan *scan) tryReadAttributes(resp *http.Response) *zgrab2.ScanError {
+	body := []byte(resp.BodyText)
+	// TODO: Cite RFC justification for this
+	// Reject successful responses which specify non-IPP MIME mediatype (ie: text/html)
+	if ipp, _ := isIPP(resp); !ipp {
+		// TODO: Log error if any
+		return zgrab2.NewScanError(zgrab2.SCAN_PROTOCOL_ERROR, errors.New("IPP Content-Type not detected."))
+	}
 
-		for _, attr := range scan.results.Attributes {
-			// TODO: Make this record all CUPS versions given. Currently records first version from first attribute.
-			if attr.Name == CupsVersion && scan.results.AttributeCUPSVersion == "" {
-				scan.results.AttributeCUPSVersion = string(attr.Values[0].Bytes)
-			}
-			// TODO: Make this report all IPP versions given. Currently records all versions from first attribute.
-			if attr.Name == VersionsSupported && len(scan.results.AttributeIPPVersions) == 0 {
-				for _, v := range attr.Values {
-					scan.results.AttributeIPPVersions = append(scan.results.AttributeIPPVersions, string(v.Bytes))
-				}
-			}
-			// TODO: Make this record all printer URI's given. Currently records the first uri for each attribute.
-			if attr.Name == PrinterURISupported && len(scan.results.AttributePrinterURIs) == 0 {
-				scan.results.AttributePrinterURIs = append(scan.results.AttributePrinterURIs, string(attr.Values[0].Bytes))
+	scan.results.Attributes = append(scan.results.Attributes, readAllAttributes(body)...)
+
+	for _, attr := range scan.results.Attributes {
+		// TODO: Make this record all CUPS versions given. Currently records first version from first attribute.
+		if attr.Name == CupsVersion && scan.results.AttributeCUPSVersion == "" {
+			scan.results.AttributeCUPSVersion = string(attr.Values[0].Bytes)
+		}
+		// TODO: Make this report all IPP versions given. Currently records all versions from first attribute.
+		if attr.Name == VersionsSupported && len(scan.results.AttributeIPPVersions) == 0 {
+			for _, v := range attr.Values {
+				scan.results.AttributeIPPVersions = append(scan.results.AttributeIPPVersions, string(v.Bytes))
 			}
 		}
+		// TODO: Make this record all printer URI's given. Currently records the first uri for each attribute.
+		if attr.Name == PrinterURISupported && len(scan.results.AttributePrinterURIs) == 0 {
+			scan.results.AttributePrinterURIs = append(scan.results.AttributePrinterURIs, string(attr.Values[0].Bytes))
+		}
 	}
+
+	return nil
 }
 
 func versionNotSupported(body string) bool {
@@ -346,13 +353,10 @@ func (scanner *Scanner) augmentWithCUPSData(scan *scan, target *zgrab2.ScanTarge
 	if cupsResp.StatusCode != 200 {
 		return zgrab2.NewScanError(zgrab2.SCAN_APPLICATION_ERROR, errors.New("Response returned with status " + cupsResp.Status))
 	}
-	// TODO: Cite RFC justification for this
-	// Reject successful responses which specify non-IPP MIME mediatype (ie: text/html)
-	if ipp, _ := isIPP(cupsResp); !ipp {
-		// TODO: Log error if any
-		return zgrab2.NewScanError(zgrab2.SCAN_PROTOCOL_ERROR, errors.New("IPP Content-Type not detected."))
+
+	if err := scan.tryReadAttributes(scan.results.CUPSResponse); err != nil {
+		return err
 	}
-	scan.tryReadAttributes(scan.results.CUPSResponse.BodyText)
 	return nil
 }
 
@@ -393,33 +397,31 @@ func sendIPPRequest(scan *scan, body *bytes.Buffer) (*http.Response, *zgrab2.Sca
 	return resp, nil
 }
 
-func hasContentType(resp *http.Response, contentType string) (bool, error) {
+func hasContentType(resp *http.Response, contentType string) bool {
 	// Removal of everything post-comma added in response to empirical examples of Virata-EmWeb
 	// print servers listed with "Content-Type" of "application/ipp, public"
 	cType := strings.Split(resp.Header.Get("Content-Type"), ",")[0]
-	// TODO: Capture parameters and report them in ScanResults?
 	// Parameters can be ignored, since there are no required or optional parameters
 	// IPP parameters specified at https://www.iana.org/assignments/media-types/application/ipp
 	mediatype, _, err := mime.ParseMediaType(cType)
-	// TODO: See if empty media type is sufficient as failure indicator,
-	// there could be other states where reading mediatype screwed up, but isn't empty (ie: corrupted/malformed)
+	// Certainly doesn't have correct Content-Type if there was a malformed or empty Content-Type
 	if mediatype == "" && err != nil {
-		//TODO: Handle errors in a weird way, since media type is still returned
-		//      if there's an error when parsing optional parameters
-		return false, err
+		return false
 	}
-	// FIXME: Maybe pass the error along, maybe not. We got what we wanted.
-	// Check for subtype alone added in resonse to empirical examples of Rapid Logic print servers
+	// Check for only subtype added in resonse to empirical examples of Rapid Logic print servers
 	// listed with "Content-Type" of "IPP"
 	subType := strings.Split(contentType, "/")[1]
-	return strings.HasPrefix(mediatype, contentType) || strings.HasPrefix(mediatype, subType), nil
+	return strings.HasPrefix(mediatype, contentType) || strings.HasPrefix(mediatype, subType)
 }
 
-// TODO: Determine whether to return error
-func isIPP(resp *http.Response) (bool, error) {
-	hasIPP, _ := hasContentType(resp, ContentType)
+func isIPP(resp *http.Response) bool {
+	hasIPP := hasContentType(resp, ContentType)
 	body := []byte(resp.BodyText)
-	return resp.StatusCode == 200 && (hasIPP || bytes.Contains(body, AttributesCharset)), nil
+	// If Content-Type header doesn't clearly indicate IPP, but "attributes-charset"
+	// attribute is specified in the correct format for IPP, still indicate a positive detection
+	// This is in response to empirical evidence of all false negatives specifying "attributes-charset"
+	// in the correct format.
+	return resp.StatusCode == 200 && (hasIPP || bytes.Contains(body, AttributesCharset))
 }
 
 func (scanner *Scanner) Grab(scan *scan, target *zgrab2.ScanTarget, version *version) *zgrab2.ScanError {
@@ -477,14 +479,10 @@ func (scanner *Scanner) Grab(scan *scan, target *zgrab2.ScanTarget, version *ver
 	if resp.StatusCode != 200 {
 		return zgrab2.NewScanError(zgrab2.SCAN_APPLICATION_ERROR, errors.New("Response returned with status " + resp.Status))
 	}
-	// TODO: Cite RFC justification for this
-	// Reject successful responses which specify non-IPP MIME mediatype (ie: text/html)
-	if ipp, _ := isIPP(resp); !ipp {
-		// TODO: Log error if any
-		return zgrab2.NewScanError(zgrab2.SCAN_PROTOCOL_ERROR, errors.New("IPP Content-Type not detected."))
-	}
 
-	scan.tryReadAttributes(scan.results.Response.BodyText)
+	if err := scan.tryReadAttributes(scan.results.Response); err != nil {
+		return err
+	}
 	if scan.results.CUPSVersion != "" {
 		err := scanner.augmentWithCUPSData(scan, target, version)
 		if err != nil {
--- a/modules/ipp/scanner.go
+++ b/modules/ipp/scanner.go
@@ -24,6 +24,7 @@ import (
 )
 
 const (
+	AttributesCharset   []byte = [0x47, 0x00, 0x12, 0x61, 0x74, 0x74, 0x72, 0x69, 0x62, 0x75, 0x74, 0x65, 0x73, 0x2d, 0x63, 0x68, 0x61, 0x72, 0x73, 0x65, 0x74]
 	ContentType         string = "application/ipp"
 	VersionsSupported   string = "ipp-versions-supported"
 	CupsVersion         string = "cups-version"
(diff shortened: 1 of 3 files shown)


modules/ipp/scanner.go
```diff
@@ -43,6 +43,7 @@ var (
 	ErrVersionNotSupported = errors.New("IPP version not supported")
 
 	Versions = []version{{Major: 2, Minor: 1}, {Major: 2, Minor: 0}, {Major: 1, Minor: 1}, {Major: 1, Minor: 0}}
+	AttributesCharset = []byte{0x47, 0x00, 0x12, 0x61, 0x74, 0x74, 0x72, 0x69, 0x62, 0x75, 0x74, 0x65, 0x73, 0x2d, 0x63, 0x68, 0x61, 0x72, 0x73, 0x65, 0x74}
```
justinbastress: Future: readability would be improved if this was something like 
```
func getTaggedString(tag byte, val string) []byte {
    // return tag | len(val) | []byte(val)
}
AttributesCharset = getTaggedString(0x47, "attributes-charset")
```

modules/ipp/scanner.go
```diff
@@ -225,80 +211,149 @@ func bufferFromBody(res *http.Response, scanner *Scanner) *bytes.Buffer {
...
+	Values []Value `json:"values,omitempty"`
+	ValueTag byte  `json:"tag,omitempty"`
+}
+
+type AttributeGroup struct {
+	Attrs []*Attribute
+	Tag byte
+}
+
+func printGroup(g *AttributeGroup) {
```
justinbastress: Unused?

modules/ipp/scanner.go
```diff
@@ -225,80 +211,149 @@ func bufferFromBody(res *http.Response, scanner *Scanner) *bytes.Buffer {
...
-				"attribute": VersionsSupported,
-			}).Debug("Failed to read attribute.")
-		} else {
-			for _, v := range ippVersions {
-				scan.results.AttributeIPPVersions = append(scan.results.AttributeIPPVersions, string(v))
+}
+
+// TODO: Log every error that could come out of this
+// TODO: Determine whether errors should be ignored, debug logged, fatal, etc.
+func readAllAttributes(body *[]byte) (*[]*Attribute) {
```
justinbastress: FYI, slices are already passed by reference; you would only need to take in a pointer if you intended to make body point to a new slice.

Also, should this return `(*[]*Attribute, error)` (or `([]*Attribute, error)`? The fact that it's in parens makes it look like maybe it used to.

modules/ipp/scanner.go
```diff
@@ -225,80 +211,149 @@ func bufferFromBody(res *http.Response, scanner *Scanner) *bytes.Buffer {
...
+	buf := bytes.NewBuffer(*body)
+	var start struct {
+		Version int16
+		StatusCode int16
+		ReqID int32
+	}
+	binary.Read(buf, binary.BigEndian, &start)
```
justinbastress: - [ ] Confirm that it was able to read all 8 bytes?

modules/ipp/scanner.go
```diff
@@ -225,80 +211,149 @@ func bufferFromBody(res *http.Response, scanner *Scanner) *bytes.Buffer {
...
+	var tag byte
+	binary.Read(buf, binary.BigEndian, &tag)
```
justinbastress: Confirm it was able to read the tag

modules/ipp/scanner.go
```diff
@@ -225,80 +211,149 @@ func bufferFromBody(res *http.Response, scanner *Scanner) *bytes.Buffer {
...
+	var groups []AttributeGroup
+	// Until encountering end-of-attributes-tag:
+	// Per-group loop
+	for tag != 0x03 {
```
justinbastress: It looks like this could be simplified / un-tabbed by checking `if tag == 3 { return empty list }`?

modules/ipp/scanner.go
```diff
@@ -225,80 +211,149 @@ func bufferFromBody(res *http.Response, scanner *Scanner) *bytes.Buffer {
...
+		// Create a new group
+		group := AttributeGroup{Tag: tag}
+		// Read the first tag:
+		binary.Read(buf, binary.BigEndian, &tag)
+		var lastTag byte
+		for tag < 0x00 || tag > 0x05 {
```
justinbastress: Note: `byte` is a `uint8`, so it's never `< 0x00`; if you need a signed byte, you can use `var tag int8`

modules/ipp/scanner.go
justinbastress: - [ ] Check `binary.Read` for errors / incomplete reads everywhere
 - [ ] Add comments detailing the format (briefly, but with a RFC link)

modules/ipp/scanner.go
```diff
@@ -225,80 +211,149 @@ func bufferFromBody(res *http.Response, scanner *Scanner) *bytes.Buffer {
...
+			// TODO: Implement parsing attribute collections (they're special)
```
justinbastress: This loop (and the function overall) should have inline comments describing how the parsing is done -- e.g. "Read the name length, if it is zero, then append the value to the previous attribute" etc etc -- a good rule of thumb would be a brief statement on each line where the input or output are touched.

modules/ipp/scanner.go
```diff
@@ -225,80 +211,149 @@ func bufferFromBody(res *http.Response, scanner *Scanner) *bytes.Buffer {
...
+				attr.Name = string(name)
+			}
+			var length int16
+			binary.Read(buf, binary.BigEndian, &length)
+			val := make([]byte, length)
+			binary.Read(buf, binary.BigEndian, &val)
+			attr.Values = append(attr.Values, Value{Bytes: val})
+
+			if len(group.Attrs) == 0 || attr != group.Attrs[len(group.Attrs)-1] {
+				group.Attrs = append(group.Attrs, attr)
```
justinbastress: Could this check be avoided by adding the attr to the group when it is created?

modules/ipp/scanner.go
```diff
@@ -225,80 +211,149 @@ func bufferFromBody(res *http.Response, scanner *Scanner) *bytes.Buffer {
...
 			}
+			lastTag = tag
+			binary.Read(buf, binary.BigEndian, &tag)
```
justinbastress: If there is no data left, `binary.Read` will return an error and leave `tag` equal to its previous value...which seems like it could lead to an infinite loop?
 - [ ] Error handling / check break conditions

modules/ipp/scanner.go
```diff
@@ -225,80 +211,149 @@ func bufferFromBody(res *http.Response, scanner *Scanner) *bytes.Buffer {
...
-			log.WithFields(log.Fields{
-				"error":     err,
-				"attribute": VersionsSupported,
-			}).Debug("Failed to read attribute.")
-		} else {
-			for _, v := range ippVersions {
-				scan.results.AttributeIPPVersions = append(scan.results.AttributeIPPVersions, string(v))
+}
+
+// TODO: Log every error that could come out of this
```
justinbastress: Some errors off the top of my head --

 * Data too short (an immediate check that `len(body) > sizeof(start) + sizeof(tag) + sizeof(len)`, or if you get an io.EOF when you're not expecting it)
 * Invalid data (e.g. if you get a length that is longer than the remaining data, or if you can't access the length of the remaining data, longer than `len(body)`)
 * Read error (the only one that may need special handling is `io.EOF`, otherwise I don't see why they couldn't just be returned as-is)

modules/ipp/scanner.go
```diff
@@ -225,80 +211,149 @@ func bufferFromBody(res *http.Response, scanner *Scanner) *bytes.Buffer {
...
+	var attrs []*Attribute
+	for _, g := range groups {
+		attrs = append(attrs, g.Attrs...)
+	}
+	return &attrs
+}
+
+func (scan *scan) tryReadAttributes(body string) {
+	bodyBytes := []byte(body)
+	if bytes.Contains(bodyBytes, []byte(AttributesCharset)) {
```
justinbastress: Is charset guaranteed to always be present in all valid responses? This probably needs a comment either giving the reference to that fact, or explaining why it is an acceptable proxy.

Also -- isn't `AttributesCharset` already a `[]byte`?

modules/ipp/scanner.go
```diff
@@ -225,80 +211,149 @@ func bufferFromBody(res *http.Response, scanner *Scanner) *bytes.Buffer {
...
+		scan.results.Attributes = append(scan.results.Attributes, *readAllAttributes(&bodyBytes)...)
+
+		for _, attr := range scan.results.Attributes {
+			// TODO: Make this record all CUPS versions given. Currently records first version from first attribute.
```
justinbastress: If having multiple CUPS versions is normal, then that would be a good idea -- but if it's a weird special case, then you are probably fine just including it once.

modules/ipp/scanner.go
```diff
@@ -225,80 +211,149 @@ func bufferFromBody(res *http.Response, scanner *Scanner) *bytes.Buffer {
...
+			if attr.Name == CupsVersion && scan.results.AttributeCUPSVersion == "" {
+				scan.results.AttributeCUPSVersion = string(attr.Values[0].Bytes)
+			}
+			// TODO: Make this report all IPP versions given. Currently records all versions from first attribute.
```
justinbastress: Same comments as above, though of course there are definitely multiple versions allowed here.

 - [ ] Support multiple `VersionsSupported`

modules/ipp/scanner.go
```diff
@@ -225,80 +211,149 @@ func bufferFromBody(res *http.Response, scanner *Scanner) *bytes.Buffer {
...
+			if attr.Name == VersionsSupported && len(scan.results.AttributeIPPVersions) == 0 {
+				for _, v := range attr.Values {
+					scan.results.AttributeIPPVersions = append(scan.results.AttributeIPPVersions, string(v.Bytes))
+				}
+			}
+			// TODO: Make this record all printer URI's given. Currently records the first uri for each attribute.
+			if attr.Name == PrinterURISupported && len(scan.results.AttributePrinterURIs) == 0 {
+				scan.results.AttributePrinterURIs = append(scan.results.AttributePrinterURIs, string(attr.Values[0].Bytes))
```
justinbastress: - [ ] Support multiple `AttributePrinterURIs`

modules/ipp/scanner.go
```diff
@@ -341,6 +396,15 @@ func (scanner *Scanner) augmentWithCUPSData(scan *scan, target *zgrab2.ScanTarge
 		return zgrab2.NewScanError(zgrab2.SCAN_APPLICATION_ERROR, ErrVersionNotSupported)
 	}
 
+	if cupsResp.StatusCode != 200 {
+		return zgrab2.NewScanError(zgrab2.SCAN_APPLICATION_ERROR, errors.New("Response returned with status " + cupsResp.Status))
```
justinbastress: Are all responses > 200 && < 400 really invalid? Or even 4xx, is it possible that you get say a 403, but it returns some attributes in the body nonetheless?

modules/ipp/scanner.go
```diff
@@ -382,9 +446,38 @@ func sendIPPRequest(scan *scan, body *bytes.Buffer) (*http.Response, *zgrab2.Sca
...
+	// listed with "Content-Type" of "IPP"
+	subType := strings.Split(contentType, "/")[1]
+	return strings.HasPrefix(mediatype, contentType) || strings.HasPrefix(mediatype, subType), nil
+}
+
+// TODO: Determine whether to return error
+func isIPP(resp *http.Response) (bool, error) {
+	hasIPP, _ := hasContentType(resp, ContentType)
+	body := []byte(resp.BodyText)
+	return resp.StatusCode == 200 && (hasIPP || bytes.Contains(body, AttributesCharset)), nil
```
justinbastress: Same comment on HTTP status code as above

modules/ipp/scanner.go
```diff
@@ -431,28 +524,28 @@ func (scanner *Scanner) Grab(scan *scan, target *zgrab2.ScanTarget, version *ver
...
-			err := scanner.augmentWithCUPSData(scan, target, version)
-			if err != nil {
-				log.WithFields(log.Fields{
-					"error": err,
-				}).Debug("Failed to augment with CUPS-get-printers request.")
-			}
 		}
 	}
 
+	if resp.StatusCode != 200 {
```
justinbastress: Same comment on status code as above, except this is more problematic, since it won't even let you get to the partial responses (will it?)

zgrab2_schemas/zgrab2/ipp.py
```diff
@@ -147,15 +147,52 @@
     "request": http_request_full
 })
 
+# TODO: Determine whether value-tag types with same underlying form should have a different name in this mapping
```
justinbastress: - [ ] Link RFC where these types are defined

zgrab2_schemas/zgrab2/ipp.py
```diff
@@ -147,15 +147,52 @@
...
+ipp_attribute_value = SubRecord({
+    "raw": Binary(),
+    "integer": Signed32BitInteger(),
+    "boolean": Boolean(),
+    "enum": String(),
+    # TODO: Determine appropriate type for octetString w/o specified format
```
justinbastress: `Binary()` is the proper type for `octetString`; if there is another value encoded inside the octet string, and we need to access the fields of that value, then that would need special handling.

zgrab2_schemas/zgrab2/ipp.py
```diff
@@ -147,15 +147,52 @@
...
+    "charset": String(),
+    "naturalLanguage": String(),
+    "mimeMediaType": String(),
+    "memberAttrName": String(),
+})
+
+ipp_attribute = SubRecord({
+    "name": String(),
+    "values": ListOf(ipp_attribute_value),
+    "tag": Binary(),
```
justinbastress: Is `tag` always an 8-bit unsigned integer? If so, the proper type would be `Unsigned8BitInteger()`

modules/ipp/scanner.go
clayzim: Yeah, I was using that to troubleshoot a bit while writing the code. I'll remove it.

modules/ipp/scanner.go
```diff
@@ -225,80 +211,149 @@ func bufferFromBody(res *http.Response, scanner *Scanner) *bytes.Buffer {
 }
 
 // FIXME: This will read the wrong section of the body if a substring matches the attribute name passed in
-// TODO: Support reading from multiple instances of the same attribute in a response
-func readAttributeFromBody(attrString string, body *[]byte) ([][]byte, error) {
+func readAttributeFromBody(attrString string, body *[]byte) ([][][]byte, error) {
```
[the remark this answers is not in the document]
clayzim: Supposed to be unused & removed.

modules/ipp/scanner.go
```diff
@@ -263,99 +218,186 @@ type Attribute struct {
...
+/* An IPP response contains the following data (as specified in RFC 8010 Section 3.1.8
+   https://tools.ietf.org/html/rfc8010#section-3.1.8)
+bytes name
+----------------------------
+2     version-number
+2     status-code
+4     request-id
+
+(0 or more instances of the following pair of fields)
+1     delimiter-tag OR value-tag
```
justinbastress: This is really helpful. I feel like the only other information I would need to eyeball a response body is how to distinguish a value-tag from a delimiter-tag.

modules/ipp/scanner.go
```diff
@@ -68,23 +70,23 @@ type ScanResults struct {
 	MinorVersion  *int8  `json:"version_minor,omitempty"`
 	VersionString string `json:"version_string,omitempty"`
 	CUPSVersion   string `json:"cups_version,omitempty"`
+
+	Attributes           []*Attribute `json:"attributes,omitempty" zgrab:"debug"`
```
justinbastress: I think these would be of interest even outside of debug mode.

modules/ipp/scanner.go
```diff
@@ -224,83 +208,196 @@ func bufferFromBody(res *http.Response, scanner *Scanner) *bytes.Buffer {
...
+
+type Attribute struct {
+	Name string    `json:"name,omitempty"`
+	Values []Value `json:"values,omitempty"`
+	ValueTag byte  `json:"tag,omitempty"`
+}
+
+func shouldReturnAttrs(length, soFar, size, upperBound int) (bool, error) {
+	if soFar + length > size {
+		if size == upperBound {
```
justinbastress: Am I reading this wrong, or is this backwards? `size == upperBound` would mean that it got truncated, so that would be the non-error case, right?

zgrab2_schemas/zgrab2/ipp.py
```diff
@@ -147,15 +147,56 @@
...
 })
 
+# TODO: Determine whether value-tag types with same underlying form should have a different name in this mapping
+# TODO: Add method to decode these values appropriately. Encoding of different tag's attribute values specified
+# in Table 7 of RFC 8010 Section 3.9 (https://tools.ietf.org/html/rfc8010#section-3.9)
+# "value-tag" values which specify the interpretation of an attribute's value:
+# From RFC 8010 Section 3.5.2 (https://tools.ietf.org/html/rfc8010#section-3.5.2)
+# Note: value-tag values are camelCase because the names are specified that way in RFC
+ipp_attribute_value = SubRecord({
+    "raw": Binary(),
```
justinbastress: At some point, it would be nice to give all of these `doc` (and maybe `example`) tags (see e.g. https://github.com/zmap/zgrab2/blob/master/zgrab2_schemas/zgrab2/mysql.py#L62).

zgrab2_schemas/zgrab2/ipp.py
```diff
@@ -147,15 +147,56 @@
     "request": http_request_full
 })
```
justinbastress: At some point, it would be nice to give all of the types `doc` tags, and maybe some `example`s too (see e.g. https://github.com/zmap/zgrab2/blob/master/zgrab2_schemas/zgrab2/mysql.py#L62).

modules/ipp/scanner.go
```diff
@@ -555,14 +674,24 @@ func (scanner *Scanner) newIPPScan(target *zgrab2.ScanTarget) *scan {
...
-	// TODO: Implement scan.Cleanup()
+func (scanner *Scanner) tryGrabForVersions(target *zgrab2.ScanTarget, versions []version, tls bool) (*scan, *zgrab2.ScanError) {
+	scan := scanner.newIPPScan(target, tls)
+	defer scan.Cleanup()
 	var err *zgrab2.ScanError
-	for i := 0; i < len(*versions); i++ {
-		err = scanner.Grab(scan, target, &(*versions)[i])
-		if err != nil && err.Err == ErrVersionNotSupported && i < len(*versions)-1 {
+	for i := 0; i < len(versions); i++ {
+		err = scanner.Grab(scan, target, &(versions)[i])
```
justinbastress: The leftover parens around `versions` make it look like it's trying to do something weird with the order of the index/address operations.

modules/ipp/scanner.go
clayzim: You're definitely correct, I just wrote it backward.

Issue #145, opened 2018-07-05
Fix TLS scanner


justinbastress:
It seems that some flags are ignored / unused (for example, `NoSNI` is unused, and even without it, it doesn't seem that SNI is working).

justinbastress:
Fixed in #146.

PR #146, opened 2018-07-05
TLS SNI fixes / API cleanup

justinbastress:
Previously, SNI was only supported with an explicit `--server-name`, which is obviously not helpful for realistic scans (see e.g. #145).

This adds new API calls that allow the `ScanTarget` (where the domain name is, when it's present) to be included when setting up the TLS connection, and updates the TLS scan module to use these them.

## How to Test

`echo "$(dig +short www.google.com),www.google.com" | cmd/zgrab2/zgrab2 --debug tls | jp data.tls.result.handshake_log.client_hello.server_name` -- should be "www.google.com"

`echo "$(dig +short www.google.com),www.google.com" | cmd/zgrab2/zgrab2 --debug tls --server-name www.fake-google.com | jp data.tls.result.handshake_log.client_hello.server_name` -- should be "www.fake-google.com"

`echo "$(dig +short www.google.com),www.google.com" | cmd/zgrab2/zgrab2 --debug tls --no-sni | jp data.tls.result.handshake_log.client_hello.server_name` -- should be null.

## Notes & Caveats

We'll still need to update all of the other scanners currently using TLS to either `target.OpenTLS()` or `tlsFlags.GetTLSConnectionForTarget()`.

## Issue Tracking

#145, #147

### DIFF
--- a/modules/tls.go
+++ b/modules/tls.go
@@ -62,20 +62,29 @@ func (s *TLSScanner) InitPerSender(senderID int) error {
 	return nil
 }
 
+// Scan opens a TCP connection to the target (default port 443), then performs
+// a TLS handshake. If the handshake gets past the ServerHello stage, the
+// handshake log is returned (along with any other TLS-related logs, such as
+// heartbleed, if enabled).
 func (s *TLSScanner) Scan(t zgrab2.ScanTarget) (zgrab2.ScanStatus, interface{}, error) {
-	tcpConn, err := t.Open(&s.config.BaseFlags)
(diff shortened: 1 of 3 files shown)


zakird:
Does it make sense to continue having the `--server-name` flag? I'm failing to find many scenarios where this is helpful.

justinbastress:
I think the only place it's useful is in a single-host scan -- which incidentally is the case that brought this up (trying to debug TLS issues on a host, like an easier-to-use `openssl s_client`)

Issue #16, opened 2017-11-13
Implement MongoDB


zakird:

jordan-wright:
I'd be happy to take a pass at this. I'm excited to see zgrab2 and I like the modular approach. This seems a lot easier to write modules for than the previous version of zgrab. 😄 

I'd be happy to take on MongoDB and other nosql databases/document stores (e.g. elasticsearch) as well as key-value stores (memcached/redis). I'll make issues for them before working on them for tracking.

I do have a quick question though: What kind of information is important to Censys? MongoDB without authentication can offer information about the host, such as database size compared to available space. It also contains info about the various collections and more. Shodan collects this information but I wasn't sure if it'd be important to Censys to collect it. It would result in more HTTP round trips, which might add overhead for mass ipv4 scans.

Once I know what information is important to y'all, I'll get to work. My plan is to work through an initial module, taking notes on how it's implemented. I'll put those notes in a blog post to encourage other participation, and we can consider if the notes might be useful in the wiki here.

dadrian:
> I'd be happy to take a pass at this. I'm excited to see zgrab2 and I like the modular approach. This seems a lot easier to write modules for than the previous version of zgrab. 😄

Awesome!

> I do have a quick question though: What kind of information is important to Censys? MongoDB without authentication can offer information about the host, such as database size compared to available space. It also contains info about the various collections and more. Shodan collects this information but I wasn't sure if it'd be important to Censys to collect it. It would result in more HTTP round trips, which might add overhead for mass ipv4 scans.

We generally adopt the approach of "collect as much information as possible without needing to open a second TCP socket or logging in". I expect performance won't be that large of an issue here, since there shouldn't be nearly as many Mongo hosts as there are HTTPS hosts. We can always tone down some collection if needed. We'll have a better sense of this once we start seeing some initial output, etc as the module is written.

Tagging @justinbastress, who I think may have also been looking into this scanner.

jordan-wright:
> "collect as much information as possible without needing to open a second TCP socket or logging in"

This is a perfectly reasonable approach. And I suppose nothing says we can't add more information later. Really, just knowing the number of mongodb hosts is a huge lift. Collection, system, etc. information can be made available later if the cycles permit it 🤷‍♂️  Getting enriched information at the app layer is great, but alluding to the "one TCP socket" approach it definitely comes with a performance cost.

[...]

PR #156, opened 2018-08-17
Add MongoDB scanner.

parkanzky:
Added mongodb support.

## How to Test

Pass IP of mongodb host to zgrab.

## Notes & Caveats

Still needs tests

## Issue Tracking

https://github.com/zmap/zgrab2/issues/16


modules/mongodb/scanner.go
```diff
@@ -0,0 +1,346 @@
...
+	return out
+}
+
+// getBuildInfoOpMsg returns a mongodb "OP" message containing query to retrieve MongoDB build info.
+func getBuildInfoOpMsg() ([]byte) {
+	// gleaned from tshark
+	section_payload, err := bson.Marshal(bson.M{ "buildinfo": 1, "$db": "admin" })
+	if err != nil {
+		// programmer error
+		log.Fatal("Invalid BSON")
```
justinbastress: It would be nice if the specific error got logged too -- e.g. `log.Fatalf("Invalid BSON: %v", err)` or `...%s", err.String())`

modules/mongodb/scanner.go
```diff
@@ -0,0 +1,346 @@
...
+// BuildEnvironment_t holds build environment information returned by scan.
+type BuildEnvironment_t struct {
+	Distmod string `bson:"distmod,omitempty"`
+	Distarch string `bson:"distarch,omitempty"`
+	Cc string `bson:"cc,omitempty"`
+	CcFlags string `bson:"ccflags,omitempty"`
+	Cxx string `bson:"cxx,omitempty"`
+	CxxFlags string `bson:"cxxflags,omitempty"`
+	LinkFlags string `bson:"linkflags,omitempty"`
+	TargetAarch string `bson:"target_arch,omitempty"`
```
justinbastress: Typo?

modules/mongodb/scanner.go
```diff
@@ -0,0 +1,346 @@
...
+	} else {
+		query = scanner.buildInfoOpMsg
+		resplen_offset = 5
+		resp_offset = 5
+	}
+
+	scan.conn.Write(query)
+	msg, err := scan.conn.ReadMsg()
+	if err != nil {
+		return zgrab2.TryGetScanStatus(err), nil, err
```
justinbastress: Is the data returned from the first packet / the fact that `getMaxWireVersion()` didn't throw sufficient to say that this is a MongoDB server?

If so, returning the error is still correct, but instead of nil, it should return a partial response object.

If OTOH random data could get this far, then returning a nil result is the right choice.

modules/mongodb/types.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+// Write writes a full message to the connection.
+func (conn *Connection) Write(data []byte) error {
+	n, err := conn.conn.Write(data)
+	if err != nil {
+		return err
+	}
+	if n != len(data) {
+		return &zgrab2.ScanError{Status: zgrab2.SCAN_CONNECTION_CLOSED, Err: nil}
```
justinbastress: There is a built-in error for short writes -- I think it's `io.ErrShortWrite`?

In the long run, this probably should be built into the common socket wrapper code.

modules/mongodb/scanner.go
```diff
@@ -0,0 +1,346 @@
...
+	CcFlags string `bson:"ccflags,omitempty"`
+	Cxx string `bson:"cxx,omitempty"`
+	CxxFlags string `bson:"cxxflags,omitempty"`
+	LinkFlags string `bson:"linkflags,omitempty"`
+	TargetAarch string `bson:"target_arch,omitempty"`
+	TargetOS string `bson:"target_os,omitempty"`
+}
+
+// Result holds the data returned by the scan
+type Result struct {
```
justinbastress: We typically use snake_case for our JSON output -- so you would need additional `json` tags, e.g. `json:"git_version"` or `json:"build_environment,omitempty"` (the `omitempty` used to prevent JSON fields with their default go values from being written; usually we use `omitempty` unless there is a specific reason not to -- like for `int`/`bool` values where `0`/`false` are meaningful)

modules/mongodb/scanner.go
parkanzky: lol this is a typo and I actually wrote the integration test to accommodate it. will fix

modules/mongodb/scanner.go
parkanzky: Good call

modules/mongodb/scanner.go
parkanzky: ack

modules/mongodb/scanner.go
parkanzky: Yes, if getMaxWireVersion() succeeds, then the server is speaking valid mongodb protocol. Should return partial response.

modules/mongodb/types.go
parkanzky: ack

PR #159, opened 2018-08-21
Interpret IPP Attribute Syntaxes

clayzim:
Makes IPP attributes human-readable instead of just base64 encoded byte strings. Determines which parsing function to apply based on attribute's value-tag field, a byte which indicates the applicable "value syntax".

## How to Test
`make integration-test`


modules/ipp/attributevalues.go
```diff
@@ -0,0 +1,288 @@
...
+				log.WithFields(log.Fields{
+					"error": err,
+					"data": val.Bytes,
+				}).Debug("Failed to interpret data with error.")
+				return
+			}
+			// Seconds East of UTC
+			// + -> East
+			// - -> West
+			secondsEast := int(t.DirectionFromUTC) * 60 * (int(t.MinutesFromUTC) + (60 * int(t.HoursFromUTC)))
```
justinbastress: Shouldn't the last term be multiplied by another 60?

clayzim: I think the answer is two-fold:
1) No.
2) My parentheses make this really unclear.
That last hours term is only multiplied by 60, then added to the minutes term. Then that sum is multiplied by 60 again to get seconds.

justinbastress: Duh, right --  I missed the nested parens.

justinbastress:
There are a few issues here -- most immediately, hyphens aren't permitted in key names, and it would be nice if this didn't conflict with the existing schema (so e.g. renaming `base64` back to `raw`).

PR #169, opened 2018-10-01
add some tighter bounds checking in MSSQL scanner

justinbastress:
...and if there is an uncaught panic, log the body that caused it.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -40,10 +40,12 @@ var (
 	DefaultSessionTimeout = 1 * time.Minute
 )
 
-var ErrReadLimitExceeded error = errors.New("read limit exceeded")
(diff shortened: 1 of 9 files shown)


justinbastress:
Also includes #165, since otherwise `io.ReadFull` can permanently block.

conn.go
```diff
@@ -80,8 +162,8 @@ func (c *TimeoutConnection) SetDeadline(deadline time.Time) error {
 	return nil
 }
 
-// GetTimeoutDialer returns a Dialer function that dials with the given timeout
-func GetTimeoutDialer(timeout time.Duration) func(string, string) (net.Conn, error) {
+// GetTimeoutDialer returns a DialFuncn that dials with the given timeout
```
parkanzky: DialFunc typo?

PR #172, opened 2018-10-05
To ensure that the HTTP connection is not blocked!

chushuai:
https://github.com/zmap/zgrab2/blob/15127f1b89dd780c7873760d00ecff18e05c51b3/modules/http/scanner.go#L281
shoud add  a line code 
ret.client.Timeout=time.Now().Add(scanner.config.Timeout)

please see the code in
https://github.com/zmap/zgrab2/blob/15127f1b89dd780c7873760d00ecff18e05c51b3/lib/http/client.go#L609


chushuai:
I tested my changes, scanned 380 million ports, and i think i am right

chushuai:
I discovered this problem by scanning port 9003

PR #178, opened 2018-10-22
Stricter bounds checking on MySQL

justinbastress:
Previously, we did sanity checking of the packet size *after* allocating / reading, potentially dumping huge data into the error string.

Now, check prior to reading, and on failure, grab some context, then fail fast.

## How to Test

`TEST_MODULES=mysql make integration-test`

### DIFF
--- a/lib/mysql/mysql.go
+++ b/lib/mysql/mysql.go
@@ -619,8 +619,10 @@ func (c *Connection) decodePacket(body []byte) (PacketInfo, error) {
 // "<first 16 bytes>...[n - 32] bytes remaining<last 16 bytes>").
 func trunc(body []byte, n int) (result string) {
 	defer func() {
-		// Failsafe -- never return more than 96 chars.
-		result = result[:96]
+		if len(result) > 96 {
+			// Failsafe -- never return more than 96 chars.
+			result = result[:96]
+		}
 	}()
 	if body == nil {
 		return "<nil>"
--- a/lib/mysql/mysql.go
+++ b/lib/mysql/mysql.go
@@ -616,6 +616,9 @@ func (c *Connection) decodePacket(body []byte) (PacketInfo, error) {
 
 // if n, _ := io.Read(body), returns a hex representation of body[:n] that is at most 96 bytes long.
 func trunc(body []byte, n int) string {
+	if body == nil {
+		return "<nil>"
+	}
 	if n > len(body) {
 		n = len(body)
 	}
--- a/lib/mysql/mysql.go
+++ b/lib/mysql/mysql.go
@@ -614,8 +614,14 @@ func (c *Connection) decodePacket(body []byte) (PacketInfo, error) {
 	}
 }
 
-// if n, _ := io.Read(body), returns a hex representation of body[:n] that is at most 96 bytes long.
-func trunc(body []byte, n int) string {
+// with n and body as if `n, _ := io.Read(body)`, trunc(body, n) returns a hex representation of
+// body[:n] that is at most 96 characters long (longer strings are returned as
+// "<first 16 bytes>...[n - 32] bytes remaining<last 16 bytes>").
+func trunc(body []byte, n int) (result string) {
+	defer func() {
+		// Failsafe -- never return more than 96 chars.
+		result = result[:96]
+	}()
 	if body == nil {
 		return "<nil>"
 	}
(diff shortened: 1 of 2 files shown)


andrewsardone:
@parkanzky: can you put some eyes on this PR? We can leave it with you and @justinbastress to merge.

parkanzky:
@andrewsardone sure

lib/mysql/mysql.go
```diff
@@ -612,31 +614,63 @@ func (c *Connection) decodePacket(body []byte) (PacketInfo, error) {
 	}
 }
 
+// if n, _ := io.Read(body), returns a hex representation of body[:n] that is at most 96 bytes long.
```
parkanzky: So is `n` here the maximum length requested?

lib/mysql/mysql.go
```diff
@@ -612,31 +614,63 @@ func (c *Connection) decodePacket(body []byte) (PacketInfo, error) {
...
 	var header [4]byte
-	n, err := reader.Read(header[:])
+	n, err := io.ReadFull(reader, header[:])
 	if err != nil {
-		return nil, fmt.Errorf("Error reading packet header: %s", err)
+		return nil, fmt.Errorf("error reading packet header: %s", err)
 	}
 	if n != 4 {
-		return nil, fmt.Errorf("Wrong number of bytes returned (got %d, expected 4)", n)
+		return nil, fmt.Errorf("wrong number of bytes returned (got %d, expected 4)", n)
```
parkanzky: nice

lib/mysql/mysql.go
```diff
@@ -612,31 +614,63 @@ func (c *Connection) decodePacket(body []byte) (PacketInfo, error) {
...
 	}
 	seq := header[3]
-	// length is actually Uint24; clear the bogus MSB before decoding
+	// length is actually uint24; clear the bogus MSB before decoding
```
parkanzky: Should this read 'packet size' now? or is length some sub-packet-size value that I'm not seeing?

lib/mysql/mysql.go
```diff
@@ -612,31 +614,63 @@ func (c *Connection) decodePacket(body []byte) (PacketInfo, error) {
...
+		return "<nil>"
+	}
+	if n > len(body) {
+		n = len(body)
+	}
+	if n < 1 {
+		return "<empty>"
+	}
+	if n < 48 {
+		return fmt.Sprintf("%x", body[:n])
```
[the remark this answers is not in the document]
parkanzky: d'oh disregard this comment, strings of bytes are 0-padded. I just tested with a single byte originally.

lib/mysql/mysql.go
justinbastress: That's poorly worded. I was trying to give a definition of `n` and `body`; `n` being the number of bytes actually read, which can be less than `len(body)` (and `len(body)` is the number of bytes requested)

lib/mysql/mysql.go
justinbastress: You're right, should be `packetSize`, not "length".

lib/mysql/mysql.go
```diff
@@ -612,31 +614,63 @@ func (c *Connection) decodePacket(body []byte) (PacketInfo, error) {
...
+	}
+	return fmt.Sprintf("%x...[%d bytes]...%x", body[:16], n - 32, body[n-16:])
+}
```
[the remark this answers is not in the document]
justinbastress: - [ ] Clarify truncation

lib/mysql/mysql.go
justinbastress: - [ ] `s/length/packetSize/`

lib/mysql/mysql.go
justinbastress: Strictly, this shouldn't be reachable, since `io.ReadFull` should return a non-`nil` `err` if it's unable to read the entire header, but it doesn't hurt.

justinbastress:
@parkanzky I think that should address your issues.

lib/mysql/mysql.go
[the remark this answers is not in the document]
parkanzky: So you don't think it's worth it to add an explicit slice at the end? My concern is that the math is kind of fragile and if someone tweaks it, it could end up coming out to longer than 96 bytes. Seems easy to at the very end just return a slice that is max len 96..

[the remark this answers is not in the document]
justinbastress: It looks like this comment is pinned to the old revision? I added [this](https://github.com/zmap/zgrab2/blob/jb/mysql-timeout/lib/mysql/mysql.go#L621) to ensure that it never goes over, even if we do somehow get a 100MB size.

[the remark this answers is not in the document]
justinbastress: But...d'oh. Probably should only do that if `len(result) > 96`.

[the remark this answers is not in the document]
parkanzky: > It looks like this comment is pinned to the old revision? I added this to ensure that it never goes over, even if we do somehow get a 100MB size.

Oh that looks great, sorry. I think I returned to a window without seeing the refresh link at the top of the page. Thanks for the edits!

PR #188, opened 2019-01-25
use a virtualenv to install zschema

justinbastress:


integration_tests/test.sh
```diff
@@ -63,6 +65,15 @@ fi
 status=0
 failures=""
 echo "Doing schema validation..."
+
+if ! [ -d "${INTEGRATION_TEST_VENV}" ]; then
```
codyprime: You are testing that `${INTEGRATION_TEST_VENV}` is not a directory here...  You probably want a positive test for a regular file here, e.g.: `if [ -f "${INTEGRATION_TEST_VENV}" ]; then`

A negative test for 'not a directory' will test true if for some reason ${INTEGRATION_TEST_VENV} gets unset (`""`), and will also pass if it is a pipe, device file, etc..

integration_tests/test.sh
```diff
@@ -29,6 +29,8 @@ set -e
 # Run from root of project
```
codyprime: I know not part of this patch in particular, but I'd recommend setting the following:
```
set -eu
set -o pipefail
```

integration_tests/test.sh
justinbastress: The idea was to only do `virtualenv` (and the `pip install`) if the dir didn't exist -- so I don't think `test -f` does what I want, either, since that also returns false when it doesn't exist.

I guess `test -x ${VENV}/bin/python` is probably safer than either.

It feels like a natural fit for the Makefile, but I left it in here in keeping with the ability to run `test.sh` on its own.

codyprime: Ah, you are right - I read the intent wrong.  Carry on :)

integration_tests/test.sh
justinbastress: `set -u` will require some refactoring, since the un-settedness of `NOSETUP` / `NOSCHEMA` etc are used to give standard default behavior, but allow special cases.

codyprime:
My approval still stands after commit id aa2d162

PR #199, opened 2019-03-29
Add custom regexp banner grabber

AnthraX1:
New Banner module that can grab banner and match against a custom regexp. It will still get the banner and create protocol error  If "--pattern" is defined and the results did not match.

options:

          --probe=            Probe to send to the server. Use triple slashes
                                    to escape, for example \\\n is literal \n
                                    (default: \n)
          --pattern=          Pattern to match, must be valid regexp.
          --max-tries=       Number of tries for timeouts and connection
                                    errors before giving up. (default: 1)

## How to Test

echo 'google.com'|./zgrab2 banner -p 80 --pattern="asfgqwg" --probe "GET / \n\n"

Should output {"domain":"google.com","data":{"banner":{"status":"protocol-error","protocol":"banner","result":{"banner":"HTTP/1.0 .... ","error":"pattern did not match"}}}

echo 'google.com'|./zgrab2 banner -p 80 --pattern="HTTP" --probe "GET / \n\n"

Should output ... "data":{"banner":{"status":"success" ....

## Notes & Caveats
Still work in progress and I haven't got time to write proper tests...

## Issue Tracking
Use issues page.

### DIFF
--- a/modules/banner/scanner.go
+++ b/modules/banner/scanner.go
@@ -5,11 +5,12 @@ package banner
 
 import (
 	"errors"
+	"fmt"
 	"io"
 	"log"
 	"net"
 	"regexp"
-	"strings"
+	"strconv"
 
 	"github.com/zmap/zgrab2"
 )
@@ -17,7 +18,7 @@ import (
 // Flags give the command-line flags for the banner module.
 type Flags struct {
 	zgrab2.BaseFlags
-	Probe    string `long:"probe" default:"\n" description:"Probe to send to the server."`
+	Probe    string `long:"probe" default:"\\n" description:"Probe to send to the server. Use triple slashes to escape, for example \\\\\\n is literal \\n" `
 	Pattern  string `long:"pattern" description:"Pattern to match, must be valid regexp."`
 	MaxTries int    `long:"max-tries" default:"1" description:"Number of tries for timeouts and connection errors before giving up."`
 }
@@ -30,6 +31,7 @@ type Module struct {
 type Scanner struct {
 	config *Flags
 	regex  *regexp.Regexp
+	probe  []byte
 }
 
 type Results struct {
@@ -96,6 +98,11 @@ func (scanner *Scanner) Init(flags zgrab2.ScanFlags) error {
 	f, _ := flags.(*Flags)
 	scanner.config = f
 	scanner.regex = regexp.MustCompile(scanner.config.Pattern)
+	probe, err := strconv.Unquote(fmt.Sprintf(`"%s"`, scanner.config.Probe))
+	if err != nil {
+		panic("Probe error")
+	}
+	scanner.probe = []byte(probe)
 	return nil
 }
 
@@ -108,6 +115,7 @@ func (scanner *Scanner) Scan(target zgrab2.ScanTarget) (zgrab2.ScanStatus, inter
 		err     error
 		readerr error
 	)
+	fmt.Printf("Probe: %v", scanner.probe)
 	for try < scanner.config.MaxTries {
 		try += 1
 		conn, err = target.Open(&scanner.config.BaseFlags)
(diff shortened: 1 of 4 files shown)


modules/banner/scanner.go
```diff
@@ -0,0 +1,152 @@
...
+		if err != nil {
+			continue
+		}
+		break
+	}
+	if err != nil {
+		return zgrab2.TryGetScanStatus(err), nil, err
+	}
+	defer conn.Close()
+	r := strings.NewReplacer(`\n`, "\n", `\r`, "\r", `\t`, "\t")
```
justinbastress: Any reason for replacing just these, and not using e.g. JSON or another standard special-characters-in-strings format? I could see wanting to send a hex string, for example.

zgrab2_schemas/zgrab2/banner.py
```diff
@@ -0,0 +1,20 @@
+# zschema sub-schema for zgrab2's banner module
+# Registers zgrab2-banner globally, and banner with the main zgrab2 schema.
+from zschema.leaves import *
+from zschema.compounds import *
+import zschema.registry
+
+import zcrypto_schemas.zcrypto as zcrypto
+from . import zgrab2
+
+# modules/cassandra.go - CassandraScanResults
```
justinbastress: Copy/paste mistake?

```suggestion
# modules/banner/scanner.go - Results
```

modules/banner/scanner.go
AnthraX1: Sending hex string would make sense. Initially I was just using it to send weird HTTP requests. The problem is refection.tag.get will convert everything to literal string. I can't seem to find a good way around it. Any suggestions?

AnthraX1: Repacling \x to non-literal in the last commit.

justinbastress: What you have is probably preferable to forcing hex input -- I was picturing something like `json.Unmarshal([]byte("\"" + scanner.config.Probe + "\""),  &scanner.config.Probe)`, so you could do e.g. `--probe="GET /\r\n\r\n"` or `--probe="GET /\u000A\u000D\u000A\u000D"`, but that may be more complicated than necessary.

The type of issue with the current implementation (as opposed to a standard escaped-string-decoding library call) is that e.g. `"\\n"` -- which one would expect to send a literal `\n` -- would here send a slash followed by a linebreak.

modules/banner/scanner.go
```diff
@@ -0,0 +1,152 @@
...
+
+	"github.com/zmap/zgrab2"
+)
+
+// Flags give the command-line flags for the banner module.
+type Flags struct {
+	zgrab2.BaseFlags
+	Probe    string `long:"probe" default:"\n" description:"Probe to send to the server."`
+	Pattern  string `long:"pattern" description:"Pattern to match, must be valid regexp."`
+	MaxTries int    `long:"maxtries" default:"1" description:"Number of tries for timeouts and connection errors before giving up."`
```
justinbastress:
```suggestion
	MaxTries int    `long:"max-tries" default:"1" description:"Number of tries for timeouts and connection errors before giving up."`
```

modules/banner/scanner.go
AnthraX1: the problem with unmarshalled json is it only accepts unicode encoded hex string such as \u0000 not \x00. I'll work on it from another angle maybe using regexp replace

AnthraX1: Alright, problem fixed. Although triple escape is a bit annoying but I like this solution much better.

AnthraX1:
merge?

modules/banner/scanner.go
pat-inthe-hat: Would it possible to send binary data as a probe request? For example, [most] RDP services can be identified by sending and receiving a specific value. `echo 0300000b06e0000000000 | xxd -r -p | nc 192.168.1.2 3389`. Listening servers respond with `\x03\x00\x00\x0b\x06\xd0\x00\x00\x124\x00`

PR #205, opened 2019-06-05
SMB: Bounds check slice index before access

codyprime:
The smb library bounds checks for a message size that is too large, but
does not check for a message size that is way too small.  Error out if
the message size is not at least as large as the ProtocolID 4-byte
preamble.

This fixes slice out of bound panics when checking the buffer for the
protID string for certain hosts.


dadrian:
It's like Heartbleed all over again!

PR #208, opened 2019-06-19
[CEN-1492] ricky/redis

rickydia:
ZGrab2 scanner module for Redis [CEN-1492].

Sends four commands by default:
- PING
- INFO
- NONEXISTENT
- QUIT

Has four Redis specific command line flags:
- `--custom-commands [PATHNAME]`
  + Pathname for a JSON/YAML file that contains extra commands to execute. WARNING: This is sent in the clear.
- `--mappings [PATHNAME]`
  + Pathname for a JSON/YAML file that contains mappings for command names.
- `--password [PASSWORD]`
  + Set a password to use to authenticate to the server. WARNING: This is sent in the clear.
- `--inline`
  + Send commands using the inline syntax
## How to Test
`make integration-test TEST_MODULES=redis`

## Issue Tracking
[[CEN-1492]](https://censysio.atlassian.net/browse/CEN-1492)


bwireman:
--password [PASSWORD] Flag seems scary to me, even if we don't use it while scanning

modules/redis/scanner.go
```diff
@@ -69,26 +74,34 @@ type Result struct {
...
-	// simple string "OK" even when authentication is required, unless the
-	// QUIT command was renamed.
-	QuitResponse string `json:"quit_response,omitempty"`
+	// Version is read from the InfoResponse (the field "server_version"), if
+	// present.
+	Version string `json:"version,omitempty"`
+
+	// OS is read from the InfoResponse (the field "os"), if present. It specifies
+	// the OS the redis server is running.
+	OS string `json:"os,omitempty"`
```
justinbastress: Good catch - any other common / useful fields in there?

modules/redis/scanner.go
```diff
@@ -152,6 +169,71 @@ func (scan *scan) Close() {
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// Initializes the command mappings
+func (scanner *Scanner) initCommands() error {
+	scanner.commandMappings = map[string]interface{}{
```
justinbastress: Any reason this is `map[string]interface{}` instead of `map[string]string`?

rickydia: That was left over from a previous attempt implementation, so I'll change it to `map[string]string`

modules/redis/scanner.go
```diff
@@ -152,6 +169,71 @@ func (scan *scan) Close() {
...
+
+	// User supplied a file for updated command mappings
+	if scanner.config.Mappings != "" {
+		var mappings map[string]string
+		err := getFileContents(scanner.config.Mappings, &mappings)
+		if err != nil {
+			return err
+		}
+		for origCommand, newCommand := range mappings {
+			scanner.commandMappings[strings.ToUpper(origCommand)] = strings.ToUpper(newCommand)
```
justinbastress: Are commands required to be uppercase? If so, that should be called out; if not, maybe mention in that the help.

rickydia: No they're not. Redis commands are case insensitive, so sending `info` is the same as sending `INFO`. Since the commands we send are all uppercase, if the mapping updates a user passes in is `"info":"renamed_info"`, I'd want the command that we send when scanning for 'INFO' to be the new mapping. I basically did this to ensure that our commands get overwritten correctly in the case where the the inputted JSON/YAML file has lowercase commands

modules/redis/scanner.go
rickydia: Another field that might be useful is `redis_mode`, which specifies what mode Redis is running on, whether that be `cluster` or `standalone`. However, other than that I'm not sure that there's much more that might be useful

modules/redis/scanner.go
```diff
@@ -228,43 +310,64 @@ func (scanner *Scanner) Scan(target zgrab2.ScanTarget) (zgrab2.ScanStatus, inter
...
+		}
+		customResponse := CustomResponse{
+			Command:   full_cmd[0],
+			Arguments: strings.Join(full_cmd[1:], " "),
+			Response:  forceToString(resp),
+		}
+		result.CustomResponses = append(result.CustomResponses, customResponse)
+	}
+	quitResponse, err := scan.SendCommand(scanner.commandMappings["QUIT"].(string))
+	if err == io.EOF && quitResponse == nil {
```
justinbastress: So this misses the case where it returns EOF *and* some data, which doesn't seem like it should be an error; i.e. `err == io.EOF` in response to a `QUIT` should be the same as `err == nil`.

modules/redis/scanner.go
```diff
@@ -228,43 +310,64 @@ func (scanner *Scanner) Scan(target zgrab2.ScanTarget) (zgrab2.ScanStatus, inter
...
+		quitResponse = NullValue
+	} else if err != nil {
```
justinbastress: So this misses the case where it returns EOF *and* some data, which doesn't seem like it should be an error; i.e. `err == io.EOF` in response to a `QUIT` should be the same as `err == nil`.

Should this instead be
```suggestion
	} else if err != nil && err != io.EOF {
```

modules/redis/scanner.go
justinbastress: That could be nice to have -- especially since we're already reading it.

modules/redis/scanner.go
justinbastress: got it. This makes perfect sense then.

modules/redis/scanner.go
rickydia: We now collect the `redis_mode`

modules/redis/scanner.go
rickydia: Fixed!

modules/redis/scanner.go
rickydia: I changed the logic so that we don't have to check whether `err == io.ERR`. Could you check the updates and let me know if that still works as intended? If not, I can use your suggestion

modules/redis/scanner.go
```diff
@@ -212,12 +335,23 @@ func (s *Scanner) Protocol() string {
 	return "redis"
 }
 
+// Converts the string to a Uint32 if possible. If not, returns 0 (the zero value of a uin32)
+func convToUint32(s string) uint32 {
+	s_64, err := strconv.ParseUint(s, 10, 32)
+	if err != nil {
+		// TODO: LEARN HOW TO LOG
```
justinbastress: Usually bad data errors like this wouldn't get logged -- either it's not bad enough to make you fail detection, but instead just leave something out of the result (so the conversion error can be inferred from the absence of that field), or it is bad enough to fail, so you return e.g. a protocol error and put your log message in the error string.

modules/redis/scanner.go
```diff
@@ -228,44 +362,97 @@ func (scanner *Scanner) Scan(target zgrab2.ScanTarget) (zgrab2.ScanStatus, inter
...
+				fieldsFound += 1
+				result.UsedMemory = convToUint32(line_prefix_suffix[1])
+			case "total_connections_received":
+				fieldsFound += 1
+				result.ConnectionsReceived = convToUint32(line_prefix_suffix[1])
+			case "total_commands_processed":
+				fieldsFound += 1
+				result.CommandsProcessed = convToUint32(line_prefix_suffix[1])
+			}
+			if fieldsWeAreLookingFor == fieldsFound {
```
justinbastress: Using a flat counter means that duplicates will be treated the same as fields you are still looking for (to take an obvious case, say there are two dozen `os:foo\r\n` followed by the rest of the expected fields)

rickydia: Yeah, I was thinking about that. The two alternatives I thought of was either to add a ton of bools and at the end have a large if statement, or to remove the final check to break early. Which solution do you think would be best?

modules/redis/scanner.go
```diff
@@ -228,44 +362,97 @@ func (scanner *Scanner) Scan(target zgrab2.ScanTarget) (zgrab2.ScanStatus, inter
...
+	if infoResponseBulk, ok := infoResponse.(BulkString); ok {
+		fieldsWeAreLookingFor := 12
+		fieldsFound := 0
 		for _, line := range strings.Split(string(infoResponseBulk), "\r\n") {
-			if strings.HasPrefix(line, "redis_version:") {
-				result.Version = strings.SplitN(line, ":", 2)[1]
+			switch line_prefix_suffix := strings.SplitN(line, ":", 2); line_prefix_suffix[0] {
+			case "redis_version":
+				fieldsFound += 1
+				result.Version = line_prefix_suffix[1]
```
justinbastress: the second argument to `SplitN` sets a maximum number of array entries, but not the minimum -- so something like
```
arch_bits: 64
os
redis_build_id: xyz abc
```

would lead to an array index out of bounds panic, since `len(line_prefix_suffix)` on the line without a colon would be 1, not 2. This worked before because it had already ensured that there was at least one colon in the string before calling `SplitN`.

zgrab2_schemas/zgrab2/redis.py
```diff
@@ -25,6 +25,22 @@
         ]),
         "quit_response": String(doc="The response to the QUIT command.", examples=["OK"]),
         "version": String(doc="The version string, read from the the info_response (if available)."),
+        "os": String(doc="The os the redis server is running, read from the the info_response (if available)."),
+        "mode": String(doc="The mode the redis server is running (standalone or cluster), read from the the info_response (if available)."),
+        "git_sha1": String(doc="The Git Sha 1 the redis server used."),
```
justinbastress:
```suggestion
        "git_sha1": String(doc="The SHA-1 Git commit hash the redis server used."),
```

zgrab2_schemas/zgrab2/redis.py
```diff
@@ -25,6 +25,22 @@
         ]),
         "quit_response": String(doc="The response to the QUIT command.", examples=["OK"]),
         "version": String(doc="The version string, read from the the info_response (if available)."),
+        "os": String(doc="The os the redis server is running, read from the the info_response (if available)."),
```
justinbastress: The docs should probably use a consistent case for `redis`

modules/redis/scanner.go
rickydia: Great catch, thank you! I added a check to see if `len(line_prefix_suffix) == 1` then I append an empty string

modules/redis/scanner.go
justinbastress: I'd just ditch the break -- it's already walking the string once to do the split, so walking the whole resulting array shouldn't noticeably hurt the performance (unless it's already bad, in which case the split itself would be the problem, not walking the resulting array).

integration_tests/redis/test.sh
```diff
@@ -15,7 +16,11 @@ for cfg in $configs; do
     CONTAINER_NAME=zgrab_redis_${cfg}
     echo "redis/test: Testing $CONTAINER_NAME"
     CONTAINER_NAME=$CONTAINER_NAME $ZGRAB_ROOT/docker-runner/docker-run.sh redis > "$ZGRAB_OUTPUT/redis/${cfg}-normal.json"
+    CONTAINER_NAME=$CONTAINER_NAME $ZGRAB_ROOT/docker-runner/docker-run.sh redis --mappings "$CONTAINER_DIR/mappings.json" > "$ZGRAB_OUTPUT/redis/${cfg}-normal-mappings.json"
+    CONTAINER_NAME=$CONTAINER_NAME $ZGRAB_ROOT/docker-runner/docker-run.sh redis --custom-commands "$CONTAINER_DIR/extra-commands.json" > "$ZGRAB_OUTPUT/redis/${cfg}-normal-extra.json"
     CONTAINER_NAME=$CONTAINER_NAME $ZGRAB_ROOT/docker-runner/docker-run.sh redis --inline > "$ZGRAB_OUTPUT/redis/${cfg}-inline.json"
+    CONTAINER_NAME=$CONTAINER_NAME $ZGRAB_ROOT/docker-runner/docker-run.sh redis --inline --mappings "$CONTAINER_DIR/mappings.yaml" > "$ZGRAB_OUTPUT/redis/${cfg}-inline-mappings.json"
+    CONTAINER_NAME=$CONTAINER_NAME $ZGRAB_ROOT/docker-runner/docker-run.sh redis --inline --custom-commands "$CONTAINER_DIR/extra-commands.yaml" > "$ZGRAB_OUTPUT/redis/${cfg}-inline-extra.json"
```
codyprime: Minor comment:  Should probably "" all `$ZGRAB_ROOT/docker-runner/docker-run.sh` calls, in case someone is a masochist and has a directory name with a space in their development path.

rickydia: Updated, though if someone has spaces in their dev path, they deserve the pain that would've come with debugging this

modules/redis/scanner.go
```diff
@@ -152,6 +210,71 @@ func (scan *scan) Close() {
...
+	}
+	return unmarshaler, nil
+}
+
+func getFileContents(file string, output interface{}) error {
+	unmarshaler, err := getUnmarshaler(file)
+	if err != nil {
+		return err
+	}
+	fileContent, err := ioutil.ReadFile(file)
```
codyprime: since `ReadFile` reads the whole file into RAM, should we check file size first, and err out if it is too large (by some arbitrary number)?

rickydia: Since these are command line flags, I assumed when adding this that we wouldn't be malicious to ourselves, so we wouldn't add files that are too obnoxiously large

integration_tests/redis/test.sh
justinbastress: The template could probably use an overhaul; fixing quoting issues, using better detection of docker images / containers (not relying on grep), and adding `set -u` all seem like obvious improvements.

modules/redis/scanner.go
```diff
@@ -228,44 +361,84 @@ func (scanner *Scanner) Scan(target zgrab2.ScanTarget) (zgrab2.ScanStatus, inter
...
-	if ok {
+	if infoResponseBulk, ok := infoResponse.(BulkString); ok {
 		for _, line := range strings.Split(string(infoResponseBulk), "\r\n") {
-			if strings.HasPrefix(line, "redis_version:") {
-				result.Version = strings.SplitN(line, ":", 2)[1]
-				break
+			line_prefix_suffix := strings.SplitN(line, ":", 2)
+			if len(line_prefix_suffix) == 1 {
+				line_prefix_suffix = append(line_prefix_suffix, "")
+			}
```
codyprime: Comment, not a request for change: Rather than needing to do an append operation here, you could make this a bit simpler and just:

```
prefix := line_prefix_suffix[0]
var suffix string
if len(line_prefix_suffix) == 1 {
   suffix = ""
} else {
   suffix = line_prefix_suffix[1]
}
```

And just switch on `prefix` and assign / convert `suffix` as appropriate.

But this isn't some critical code path for performance, so no need to change it to optimize away a single `append` call (and since the append almost certainly won't need to reallocate for a single `""`, it probably only saves a function call or so).

modules/redis/scanner.go
justinbastress: Tend to agree, any user input is probably trusted, the exception being anything that could wreak havoc on the hosts we scan or on our scanning infrastructure

codyprime: Generally speaking, that goes against traditional defensive programming practices -- all user input is untrusted, including files.  The mistake we programmers often make is thinking "no one will use this program differently than it is used now", or to always assume the filesystem can be trusted (or that mistakes don't happen!).

rickydia: I can add the check -- what would be a good limit for the file size?

codyprime: I guess the "custom commands" file would be the larger of the two you are reading... how many custom commands are you traditionally expecting?  My guess is pretty low, so maybe an order or two of magnitude higher than that (100k?  Would you expect custom commands to ever get close to that?).

modules/redis/scanner.go
rickydia: Changed to this, however slightly changed the logic since a string's zero value is an empty string

modules/redis/scanner.go
rickydia: I'd say 100KB is a pretty large upper limit -- definitely at least a magnitude or two above what I would expect.

rickydia: That being said, I made the upper limit 100KB and added a command-line flag to raise or lower that limit

rickydia:
Are there any more changes you all feel need to be done before merging? If not, could either @codyprime or @justinbastress approve the PR so that it can be merged

zgrab2_schemas/zgrab2/redis.py
```diff
@@ -25,6 +25,22 @@
         ]),
         "quit_response": String(doc="The response to the QUIT command.", examples=["OK"]),
         "version": String(doc="The version string, read from the the info_response (if available)."),
```
justinbastress: It may be helpful to also have `version` broken out into major/minor/micro ints (assuming the above `redis_version:4.0.7` example is representative) -- `String` should let you do wildcard searching, it can be extremely painful to do int range searches using string operations.

Not a blocker, since you can always new fields later, and I think in any case we want to keep `version` as a `String` (e.g. for CVE matching).

rickydia: Just to clarify, I should have both `redis_version` as a string, and `redis_major`, `redis_minor`, and `redis_micro` as ints?

justinbastress: Yeah, something like that -- probably matching whatever nomenclature they use for the segments (major/minor/revision or patch or whatever). The string will let you catch weird stuff like "1.04-beta" or "PRERELEASE-1.54.0" or "\r\nContent-Type: application/octet-stream", but the ints will help if you need to actually tract down `1.5.13 < version < 1.6.2`

rickydia: How would I go about differentiating between having a `0` within the version and not having the information? For example, given the version `5.0.0`, both the `minor` and `patchlevel` would not appear since `0` is the zero value of an int

justinbastress: You could go the route of having the ints be pointers, so you can distinguish absent from 0, or just use 0 and add a note saying that if you do need to distinguish those, you need to look at the string; but my feeling would be that unless you are seeing a significant number of hosts with missing parts, the latter would be fine.

rickydia: from my 1%, it seems like >60% of the hosts have at least one section that's a 0

justinbastress: That would be fine, as long as (say) < 1% of them are _missing_ a section.

I guess I should have clarified that in that case, you would not be using `omitempty`, since `0` is meaningful there.

rickydia: Of the redis servers that are found, most are hidden behind auth. So not using `omitempty` would mean that on those scenarios where we get a response but aren't able to gather any more information on them, we will be storing the major, minor, and patchlevel as 0's instead of omitting them. So I decided to take your initial idea of having the ints be pointers!

PR #211, opened 2019-07-03
ricky/redis

rickydia:
Removed the tests that test the `--mappings` and `--custom-commands` flags for the Redis scanner.

## How to Test
```
make -C docker-runner clean
TEST_MODULES=redis make integration-test
```

### DIFF
--- a/integration_tests/redis/test.sh
+++ b/integration_tests/redis/test.sh
@@ -4,7 +4,11 @@ set -e
 MODULE_DIR=$(dirname $0)
 ZGRAB_ROOT=$MODULE_DIR/../..
 ZGRAB_OUTPUT=$ZGRAB_ROOT/zgrab-output
-CONTAINER_DIR="./integration_tests/redis/container"
+
+# FIXME: Find a way to mount host to container and have it pass CircleCI tests.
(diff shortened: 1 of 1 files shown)


docker-runner/docker-run.sh
```diff
@@ -7,4 +7,11 @@
 : "${CONTAINER_NAME:?}"
 
 set -e
-docker run --rm --link $CONTAINER_NAME:target -e ZGRAB_TARGET=target zgrab2_runner $@
+
+MOUNT=
```
justinbastress:
```suggestion
MOUNT=""
```
just so it's unambiguous.

docker-runner/docker-run.sh
```diff
@@ -7,4 +7,11 @@
...
+if [ -n "$MOUNT_HOST" ]; then
```
justinbastress: I feel like if we are going to give the mount special treatment (rather than having an opaque e.g. `EXTRA_DOCKER_ARGS`), we should be more opinionated; like maybe only let them set the working directory instead of explicitly taking both the host and container dirs (e.g. `-v "${DOCKER_INPUT_DIR:-"$(realpath .)"}":/var/tmp/scan-root`, and then update the entrypoint.sh to `cd /var/tmp/scan-root` before running zgrab2; then you could even have default behavior that mounts the host cwd as the container cwd, so that relative paths remain the same...

PR #218, opened 2019-09-06
fix schema in redis for nonexistent_response being required

bwireman:
Fixes Small bug in the zschema for the Redis scanner, which made the field nonexistent_response required instead of nullable when generating BQ schema


justinbastress:
Worst kind of schema error.

PR #219, opened 2019-09-13
Remove ssh.key_exchange.server_host_key

dadrian:
This field doesn't exist. It was invented by a writer. Instead, there is
`ssh.server_host_key`.

## Notes & Caveats

As far as I can tell, this field doesn't exist, but a sanity check would be nice.

I dug through a bunch of scan data and don't see this field ever being populated.


justinbastress:
This one definitely does exist -- the integration test uses a pretty vanilla OpenSSH setup.

PR #227, opened 2019-10-17
Fix the formatting of IPv6 addresses

chronoslynx:
This fixes `ScanTarget.Host()` and the `host == ""` check in the HTTP module to properly encode literal IPv6 addresses as per the [scan.Dial docs](https://golang.org/pkg/net/#Dial)

## How to Test

I'm working through the integration testing framework right now to figure out how to allow it to hit the IPv6 address of the `target` container
## Notes & Caveats

We use Zgrab2 and noticed consistent issues when targeting IPv6 addresses like:
```
dial tcp: lookup AAAA:BBBB:CCC:DDD:: no such host
parse http://AAAA:BBBB::CCCC:DDDD: invalid port ":DDDD" after host
```

According to the docs for [net.Dial](https://golang.org/pkg/net/#Dial) literal IPv6 addresses must be enclosed in square brackets as in "[2001:db8::1]:80" or "[fe80::1%zone]:80".

## Issue Tracking
N/A

### DIFF
--- a/modules/http/scanner.go
+++ b/modules/http/scanner.go
@@ -11,6 +11,7 @@ import (
 	"context"
 	"crypto/sha256"
 	"errors"
+	"fmt"
 	"io"
 	"net"
 	"net/url"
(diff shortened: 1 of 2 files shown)


chronoslynx:
I'm going to close this until it's properly tested

PR #228, opened 2019-10-21
prevent running key exchange if hello-only flag set

thegwan:
Fixed hello-only bug to prevent key exchange sequence from entering at all if the hello-only flag is set

### DIFF
--- a/lib/ssh/handshake.go
+++ b/lib/ssh/handshake.go
@@ -180,29 +180,35 @@ func (t *handshakeTransport) readOnePacket() ([]byte, error) {
 	t.mu.Lock()
 
 	firstKex := t.sessionID == nil
+	if t.config.HelloOnly {
+		t.sentInitMsg = nil
+		t.sentInitPacket = nil
+		t.cond.Broadcast()
+		t.writtenSinceKex = 0
+		t.mu.Unlock()
+	} else {
+		err = t.enterKeyExchangeLocked(p)
+		if err != nil {
+			// drop connection
+			t.conn.Close()
+			t.writeError = err
+		}
 
-	err = t.enterKeyExchangeLocked(p)
-	if err != nil {
-		// drop connection
-		t.conn.Close()
-		t.writeError = err
+		if debugHandshake {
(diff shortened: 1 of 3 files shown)


thegwan:
It looks like the continuous integration test is failing on `cannot find package "github.com/cespare/xxhash/v2"`, but I don't think I'm using this?

dadrian:
@thegwan if you cherry-pick https://github.com/zmap/zgrab2/pull/229/commits/8e713fabd8d94981b4b967fc9275dc1603d4ca3c onto this PR, it should fix CI

lib/ssh/handshake.go
```diff
@@ -180,29 +180,35 @@ func (t *handshakeTransport) readOnePacket() ([]byte, error) {
 	t.mu.Lock()
 
 	firstKex := t.sessionID == nil
+	if t.config.HelloOnly {
+		t.sentInitMsg = nil
+		t.sentInitPacket = nil
+		t.cond.Broadcast()
+		t.writtenSinceKex = 0
+		t.mu.Unlock()
```
codyprime: The code looks correct, but my only concern is the duplication of the cleanup to unblock writers, and the mutex lock taken before a conditional and then unlocked in the if/else branches of the conditional.

Could the same thing be accomplished by inverting the HelloOnly check, like so:

```go
	t.mu.Lock()

	firstKex := t.sessionID == nil
	if !t.config.HelloOnly {
		err = t.enterKeyExchangeLocked(p)
		if err != nil {
			// drop connection
			t.conn.Close()
			t.writeError = err
		}

		if debugHandshake {
			log.Printf("%s exited key exchange (first %v), err %v", t.id(), firstKex, err)
		}
	}
	// Unblock writers.
	t.sentInitMsg = nil
	t.sentInitPacket = nil
	t.cond.Broadcast()
	t.writtenSinceKex = 0
	t.mu.Unlock()

	if err != nil {
		return nil, err
	}
```

This does create an unnecessary err check at the end iif HelloOnly is true, but reduces code duplication and keeps the mutex lock/unlock balanced outside conditionals.

thegwan: sounds good, making the change

PR #233, opened 2019-11-19
Honor port override when composing URL

codyprime:
Commit a38194a added an optional port override as part of the
scan target.  The HTTP and IPP modules, however, still compose
the URL (and select http vs https) by ignoring the override.

This checks for the override, and if present uses the scan target
port.  Otherwise, it falls back to the config port.

### Verify existing behavior works (no regression):

A) `echo hostname | zgrab2 http --use-https -p 443`
B) `echo hostname | zgrab2 http --use-https`

Run `A` should end in success, and also grab the TLS cert.
Run `B` should end in error, as it attempted against the default port 80 (assuming of course https is not running on this port on the host).

### Verify new behavior

This is a bit trickier, you need a different framework other than `cmd/zgrab2` that initializes the module, and performs a scan.  However, if that is done, then if the `ScanTarget` contains a port number, it should override both the default and cmdline passes ports, and use the port specified in `ScanTarget`

### DIFF
--- a/modules/http/scanner.go
+++ b/modules/http/scanner.go
@@ -284,7 +284,14 @@ func (scanner *Scanner) newHTTPScan(t *zgrab2.ScanTarget) *scan {
 	if host == "" {
 		host = t.IP.String()
 	}
-	ret.url = getHTTPURL(scanner.config.UseHTTPS, host, uint16(scanner.config.BaseFlags.Port), scanner.config.Endpoint)
+	// Scanner Target port overrides config flag port
+	var port uint16
+	if t.Port != nil {
+		port = uint16(*t.Port)
+	} else {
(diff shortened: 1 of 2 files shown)


dadrian:
Does this still handle redirects correctly?

codyprime:
> Does this still handle redirects correctly?

It should not have any effect on redirects, it just affects the port used when composing the initial request.

I verified a redirected domain by:
```bash
$ echo subdomain.domain.tld | ./zgrab2 http 
[...]
"error":"Too many redirects"
"http":{"successes":0,"failures":1}}
$  echo subdomain.domain.tld | ./zgrab2 http --max-redirects=1
[...]
"http":{"successes":1,"failures":0}}
```

Issue #234, opened 2019-11-23
Input Format doesn't run correctly


hundan2020:
`echo 1.0.0.1, baidu.com | zgrab2 http`

by document, it should connect with the ip 1.0.0.1, but it still return the content of baidu.com, following is the response

```
{"ip":"1.0.0.1","domain":"baidu.com","data":{"http":{"status":"success","protocol":"http","result":{"response":{"status_line":"200 OK","status_code":200,"protocol":{"name":"HTTP/1.1","major":1,"minor":1},"headers":{"accept_ranges":["bytes"],"cache_control":["max-age=86400"],"connection":["Keep-Alive"],"content_length":["81"],"content_type":["text/html"],"date":["Sat, 23 Nov 2019 10:25:54 GMT"],"etag":["\"51-47cf7e6ee8400\""],"expires":["Sun, 24 Nov 2019 10:25:54 GMT"],"last_modified":["Tue, 12 Jan 2010 13:48:00 GMT"],"server":["Apache"]},"body":"\u003chtml\u003e\n\u003cmeta http-equiv=\"refresh\" content=\"0;url=http://www.baidu.com/\"\u003e\n\u003c/html\u003e\n","body_sha256":"1d2d898b8ee2611eeed02868e27b1e3db2d266b4016b54160f128187961431fb","content_length":81,"request":{"url":{"scheme":"http","host":"baidu.com","path":"/"},"method":"GET","headers":{"accept":["*/*"],"user_agent":["Mozilla/5.0 zgrab/0.x"]},"host":"baidu.com"}}},"timestamp":"2019-11-23T18:27:24+08:00"}}}
```

codyprime:
Fixed in #240

PR #239, opened 2020-01-27
make http(s) request connect to given ip

hundan2020:
make http(s) request connect to given ip

## How to Test

run a port listener: `socat tcp-l:80 -`
run the go run command `echo 127.0.0.1,www.google.com | go run main.go http`, if listener got like following, then it fixed

```
GET / HTTP/1.1
Host: www.google.com
User-Agent: Mozilla/5.0 zgrab/0.x
Accept: */*
Accept-Encoding: gzip
```

## Notes & Caveats

these code didn't alter core code but just add a little compatible for this issue.

## Issue Tracking

https://github.com/zmap/zgrab2/issues/234

### DIFF
--- a/lib/http/transport.go
+++ b/lib/http/transport.go
@@ -582,12 +582,23 @@ func (e *envOnce) reset() {
 	e.val = ""
 }
 
+
 func (t *Transport) connectMethodForRequest(treq *transportRequest) (cm connectMethod, err error) {
 	if port := treq.URL.Port(); !validPort(port) {
 		return cm, fmt.Errorf("invalid URL port %q", port)
 	}
 	cm.targetScheme = treq.URL.Scheme
-	cm.targetAddr = canonicalAddr(treq.URL)
+
+	if treq.Request.RemoteAddr != "" {
+		port := treq.URL.Port()
+		if port == "" {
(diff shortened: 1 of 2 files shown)


codyprime:
Thanks for the PR!  Coincidentally, I've been working on addressing this same issue.

However, I don't think the approach in this PR is correct; we want to leave the domain name in the requesting URL, as it may be parsed by the server.  Also, if possible we want to avoid modifying the `lib/http` code, as it makes upgrades to match the upstream golang `http` library more difficult.

I think the proper approach is a custom (fake) DNS Resolver, which go 1.9 added support for by allowing the `Resolver.Dial` func to be overwritten (https://golang.org/pkg/net/#Resolver).  This approach can be implemented without modifying the `lib/http` code, as well.

I just pushed my work for this in a PR, here: https://github.com/zmap/zgrab2/pull/240

codyprime:
Another PR approach, that uses the `Hosts:` field in the HTTP Request header: https://github.com/zmap/zgrab2/pull/241

codyprime:
Thanks again for the PR.  I think this issue has been fixed in PR#241, which is now merged.  In light of that, I'll go ahead and close this PR.

PR #240, opened 2020-01-30
Add the ability to use a custom (fake) DNS resolver

codyprime:
When performing HTTP scans with a domain name, we want to be able to scan that domain name with the specified IP address.  Currently, if a domain is specified, the IP address is ignored and we scan whatever the name resolves to.

This adds a fake resolver, that can be attached to a Dialer and cause the http connection to always return the IP address we  specify.  In the scanner, this also checks that the intended domain name we are about to connect to in our dialContext matches the original target domain name; if it does not, we do not attach the custom/fake resolver.

## How to Test

```
echo 172.217.9.78,google.com | ./zgrab2 http --port 443 --use-https
```
This should return valid results.  Replace the IP address with something like 127.0.0.1:
```
echo 127.0.0.1,google.com | ./zgrab2 http --port 443 --use-https
```
Assuming no web server is running there, it will return a failure.

### DIFF
--- a/fake_resolver.go
+++ b/fake_resolver.go
@@ -1,7 +1,3 @@
-// Fake DNS lookups, so that they always returned the specified
-// IP
-//
-// Inspired by the golang net/dnsclient_unix_test.go code
 package zgrab2
 
 import (
@@ -13,6 +9,9 @@ import (
 	"time"
 )
 
+// Fake DNS Resolver, to force a DNS lookup to return a pinned address
+// Inspired by the golang net/dnsclient_unix_test.go code
+//
 // For a given IP, create a new Resolver that wraps a fake
 // DNS server. This resolver will always return an IP that
 // is represented by "ipstr", for DNS queries of the same
--- a/fake_resolver.go
+++ b/fake_resolver.go
@@ -1,0 +1,166 @@
+// Fake DNS lookups, so that they always returned the specified
+// IP
+//
+// Inspired by the golang net/dnsclient_unix_test.go code
+package zgrab2
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"golang.org/x/net/dns/dnsmessage"
+	"net"
+	"time"
+)
+
+// For a given IP, create a new Resolver that wraps a fake
+// DNS server. This resolver will always return an IP that
+// is represented by "ipstr", for DNS queries of the same
+// IP type.  Otherwise, it will return a DNS lookup error.
+func NewFakeResolver(ipstr string) (*net.Resolver, error) {
+	ip := net.ParseIP(ipstr)
+	if len(ip) < 4 {
+		return nil, fmt.Errorf("Fake resolver can't use non-IP '%s'", ipstr)
+	}
+	fDNS := FakeDNSServer{
+		IP: ip,
+	}
+	return &net.Resolver{
+		PreferGo: true, // Needed to force the use of the Go internal resolver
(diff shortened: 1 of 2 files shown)


codyprime:
Added `do-no-merge` tag because this can be done cleaner via PR #241.  I would move it to draft, but that doesn't seem possible.  We may want patch #1 for other functionality in the future, but patch #2 can be dropped.

modules/http/scanner.go
```diff
@@ -161,6 +162,25 @@ func (scan *scan) withDeadlineContext(ctx context.Context) context.Context {
...
+	// don't want to just let the http library resolve the domain.  Create a
+	// fake resolver that we will use, that always returns the IP we are given
+	// to scan.
+	if scan.target.IP != nil && scan.target.Domain != "" {
+		// In the case of redirects, we don't want to blindly use the IP we
+		// were given to scan, however.  So only use the fake resolver if the
+		// domain originally specified for the scan target machines the current
+		// address being looked up in this DialContext.
+		targetAddr := fmt.Sprintf("%s:%d", scan.target.Domain, scan.getPort())
+		if addr == targetAddr {
```
justinbastress: One exception to this exception -- what about a redirect to the same host, but a different port (80-443 for example)? It feels like it could just compare the host part of the addr, but I haven't gamed out what other weird cases that might open up.

codyprime:
Some wireshark captures:

Command: `echo 172.217.0.14,google.com | ./zgrab2-jcody http --port 80 --max-redirects=3`
image

Command: `echo 172.217.0.14,example.com | ./zgrab2-jcody http --port 80`
image

modules/http/scanner.go
```diff
@@ -158,12 +158,40 @@ func (scan *scan) withDeadlineContext(ctx context.Context) context.Context {
...
+					resolver, err := zgrab2.NewFakeResolver(scan.target.IP.String())
+					if err != nil {
+						return nil, err
+					}
+					dialer.Dialer.Resolver = resolver
+				}
+			}
+		}
+	}
+
```
codyprime: This hunk (lines 164-181) addresses the concerns @justinbastress raised, and also adds an additional check against the specified `network` (similar to what `internetAddrList` does in the go net library)

modules/http/scanner.go
codyprime: Resolved, see https://github.com/zmap/zgrab2/pull/240#discussion_r373658429

modules/http/scanner.go
justinbastress: I think I follow -- this `dialContext` gets called by the HTTP lib for all redirects, and it can decide to pass in the `addr` as a `hostname(?!):port`; if the redirect is to our "pinned" host, set a custom resolver for this connection only (a subsequent redirect leads to another `dialContext`, another `Dialer`, etc). If it's to another host, we use the default resolver (but if that host redirects back to our pinned host, we are again by another call to this).

codyprime: Yep!

fake_resolver.go
```diff
@@ -0,0 +1,166 @@
+// Fake DNS lookups, so that they always returned the specified
+// IP
+//
+// Inspired by the golang net/dnsclient_unix_test.go code
```
dadrian: This is going to show up as package-level GoDoc, not file level. You might want to move this comment to be on a struct or something.

dadrian:
I like this, and I think it's better than #241

PR #241, opened 2020-01-30
Use host header, and always connect to IP

codyprime:
This is an alternate approach to PR #240 and PR #239, that forces use of the host header, while always preferring the IP address for the host.

The only drawback is when zgrab2 is populating the HTTP Request, the ip address will be used as the host in the URL.  Whether or not this is important will depend on what is parsing the results, I think.  
 
According to the go documentation, overwriting the Request.Host field is supported and will be honored when executing:

From https://golang.org/src/net/http/request.go?#L233:
```go
    // For client requests, Host optionally overrides the Host
    // header to send. If empty, the Request.Write method uses
    // the value of URL.Host. Host may contain an international
    // domain name.
    Host string
```

## How to Test

```
echo 172.217.9.78,google.com | ./zgrab2 http --port 443 --use-https
```
This should return valid results.  Replace the IP address with something like 127.0.0.1:
```
echo 127.0.0.1,google.com | ./zgrab2 http --port 443 --use-https
```
Assuming no web server is running there, it will return a failure.

## Notes & Caveats

The HTTP Request struct will have the ip address in the URL field, instead of the domain name.
If we make proxy requests in the future, the absolute-uri will not contain the domain name, and we may want something like the FakeResolver from PR #240.

### DIFF
--- a/modules/http/scanner.go
+++ b/modules/http/scanner.go
@@ -280,9 +280,9 @@ func (scanner *Scanner) newHTTPScan(t *zgrab2.ScanTarget) *scan {
 	ret.client.Transport = ret.transport
 	ret.client.Jar = nil // Don't send or receive cookies (otherwise use CookieJar)
 	ret.client.Timeout = scanner.config.Timeout
-	host := t.Domain
-	if host == "" {
-		host = t.IP.String()
+	host := t.IP.String()
+	if t.IP == nil {
+		host = t.Domain
 	}
 	// Scanner Target port overrides config flag port
 	var port uint16
@@ -305,6 +305,9 @@ func (scan *scan) Grab() *zgrab2.ScanError {
 	}
 	// TODO: Headers from input?
 	request.Header.Set("Accept", "*/*")
+	if scan.target.Domain != "" {
+		request.Host = scan.target.Domain
+	}
 	resp, err := scan.client.Do(request)
 	if resp != nil && resp.Body != nil {
 		defer resp.Body.Close()


dadrian:
@codyprime Can you screencap a wireshark of this connecting to Google with the HTTP details broken out?

codyprime:
> @codyprime Can you screencap a wireshark of this connecting to Google with the HTTP details broken out?

From: 
`echo 172.217.0.14,google.com | ./zgrab2-jcody http --port 80`

image

And then with:
`echo 172.217.0.14,example.com | ./zgrab2-jcody http --port 80`

image

dadrian:
Maybe I'm overthinking of this, but the URL is only used for DNS resolution by a client. Once you know what IP you are connecting to, either because you specified it was in the URL or you resolved a hostname in the URL, once you set the `Host` header to a name, there's no difference on the wire between what would be sent if you went to a URL.

Specifically, connecting to `http://1.2.3.4` and setting `Host: example.com` is exactly the same as connecting to `http://example.com` except force-resolving `example.com` to 1.2.3.4.

So what URL are you referring to when you say this:

> The only drawback is when zgrab2 is populating the HTTP Request, the ip address will be used as the host in the URL.

codyprime:
> Maybe I'm overthinking of this, but the URL is only used for DNS resolution by a client. Once you know what IP you are connecting to, either because you specified it in the URL or you resolved a hostname in the URL, once you set the `Host` header to a name, there's no difference on the wire between a URL with an IP where I override `Host`, and what would be sent if you went to a URL with a hostname in it.
> 
> Specifically, connecting to `http://1.2.3.4` and setting `Host: example.com` is exactly the same as connecting to `http://example.com` except force-resolving `example.com` to 1.2.3.4.
> 
> So what URL are you referring to when you say this:
> 
> > The only drawback is when zgrab2 is populating the HTTP Request, the ip address will be used as the host in the URL.

I meant in the zgrab2 results struct that is populated and returned back.  Programs parsing that output may be expecting it to be the requested domain, since that was the old behavior of zgrab2.

The only time I could see the URL going out over the wire would be in the case of acting like a proxy (e.g. using `Request.WriteProxy()`), where the Request-line has an absolute-uri.  Zgrab2 currently does not do that, but if we did in the future then this approach could be problematic and we may require something like a FakeResolver in that case.

modules/http/scanner.go
```diff
@@ -280,9 +280,9 @@ func (scanner *Scanner) newHTTPScan(t *zgrab2.ScanTarget) *scan {
 	ret.client.Transport = ret.transport
 	ret.client.Jar = nil // Don't send or receive cookies (otherwise use CookieJar)
 	ret.client.Timeout = scanner.config.Timeout
-	host := t.Domain
-	if host == "" {
-		host = t.IP.String()
+	host := t.IP.String()
```
dadrian: Isn't this going to crash now if `t.IP` is nil?

codyprime: No, in that case IP.String() actually returns the string `"<nil>"`

codyprime:
> What happens if there is a redirect to a different path on the same host in this case? E.g. I do something like `--max-redirects=2 --path=/` and then connected to `example.com,1.2.3.4`, where 1.2.3.4 is a host that redirects to `example.com/home.php` or similar?

In that case the http lib is going to do a DNS resolution on the redirect URL, unless that URL is also IP based... so it may not be the same ip address you specified.

codyprime:
> > What happens if there is a redirect to a different path on the same host in this case? E.g. I do something like `--max-redirects=2 --path=/` and then connected to `example.com,1.2.3.4`, where 1.2.3.4 is a host that redirects to `example.com/home.php` or similar?
> 
> In that case the http lib is going to do a DNS resolution on the redirect URL, unless that URL is also IP based... so it may not be the same ip address you specified.

One way around that would be to handle redirects manually, as fully separate scans, so that you could again specify an ip/name combo.  This is also a use case where the approach in PR #240 could be used, so that the FakeResolver was used for all lookups that matched the domain.

codyprime:
In light of https://github.com/zmap/zgrab2/pull/240#pullrequestreview-351150078, I'm marking this as `do-not-merge`.  Once PR #240 is merged in, I'll close this one.

PR #244, opened 2020-02-10
Add option to flush output after each line of output

cablej:
This PR adds an option to flush output after each line of output. This may be desirable in streaming applications, where a program continually ingests zgrab2's output.

## How to Test

Run zgrab2 with the flag `--flush`. Observe that the output will be flushed after each line, rather than at the end.

### DIFF
--- a/config.go
+++ b/config.go
@@ -20,6 +20,7 @@ type Config struct {
 	LocalAddress       string          `long:"source-ip" description:"Local source IP address to use for making connections"`
 	Senders            int             `short:"s" long:"senders" default:"1000" description:"Number of send goroutines to use"`
 	Debug              bool            `long:"debug" description:"Include debug fields in the output."`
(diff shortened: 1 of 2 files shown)


luciddr34m3r:
Will this be merged?

SoloSteve:
I would also appreciate this feature

dadrian (approved):

fbexiga:
This change broke the build, did no one test this first?

dadrian:
> This change broke the build, did no one test this first?

This comment broke my graciousness, did no one think first?

frbexiga:
I apologize if my previous comment violated some guidelines, I removed it since it has. As a contributor (one-time but still) and a user of these tools myself, I was trying to alert of the issue, and didn't submit a PR myself because I'm not familiar with the code enough to know if I wouldn't break it again. I didn't mean anything else by it.

That said, I think blocking me from the org was kind of a drastic measure. Would you mind unblocking me?

PR #246, opened 2020-02-14
wip: bug: use source address for real

dadrian:
The root issue is `--source-ip` is not respected.

I updated the CLI parsing code to handle source IP ranges, similar to how we handle them in ZMap. In an ideal world, this would work with ranges and CIDRs for IPv4 and IPv6. I only implemented small ranges for IPv4, to make my life easier.

The tricky part is actually getting ZGrab2 to use one of the addresses.

For starters, the config state is global-ish, so I'm wary of just reading into `config.go:config`, since if ZGrab is imported as a library, who knows what would happen.

So I hooked source IP as optional into BaseFlags, which means you would be able to specify it on a per-scan-module basis. This would probably mostly work if I can get `BaseFlags::AfterParse` wired up correctly. Unfortunately, not all connections in ZGrab2 use the `ScanTarget::Open(BaseFlags)` API. Some use a different custom `Dialer`, from `dialer.go` in ZGrab2, to set timeouts. These don't have access to base flags.

The proper fix is probably to explicitly expose a custom dialer to each ZGrab scan module. This dialer should be prepopulated based on a non-global set of settings passed as either CLI args to ZGrab, args to the specific scan module, or as structs in code using this as a library. This is vague and complicated enough I'm not even sure what I said makes sense to anyone except me.

@codyprime, any ideas on how to not refactor everything here?

## How to Test

On the plus side, lots of tests for the CLI parsing!

## Notes & Caveats

## Issue Tracking

### DIFF
--- a/config.go
+++ b/config.go
@@ -17,7 +17,6 @@ type Config struct {
 	InputFileName      string          `short:"f" long:"input-file" default:"-" description:"Input filename, use - for stdin"`
 	MetaFileName       string          `short:"m" long:"metadata-file" default:"-" description:"Metadata filename, use - for stderr"`
 	LogFileName        string          `short:"l" long:"log-file" default:"-" description:"Log filename, use - for stderr"`
-	LocalAddress       string          `long:"source-ip" description:"Local source IP address to use for making connections"`
 	Senders            int             `short:"s" long:"senders" default:"1000" description:"Number of send goroutines to use"`
 	Debug              bool            `long:"debug" description:"Include debug fields in the output."`
 	GOMAXPROCS         int             `long:"gomaxprocs" default:"0" description:"Set GOMAXPROCS"`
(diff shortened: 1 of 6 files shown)


codyprime:
> The proper fix is probably to explicitly expose a custom dialer to each ZGrab scan module. This dialer should be prepopulated based on a non-global set of settings passed as either CLI args to ZGrab, args to the specific scan module, or as structs in code using this as a library. This is vague and complicated enough I'm not even sure what I said makes sense to anyone except me.

In my opinion, although this is the path of most resistance, I think the proper implementation is what you mention in the latter part: the framework using the modules as a library should be able to provide the Dialer, so it can determine how it wants to rotate / randomly use IPs.

This will also make it more necessary for scanner modules to make sure they are using the provided Dialer, and especially being mindful of other library calls.  Would that make it too fragile?  It is probably something we'd have to rely on catching in code reviews, and hopefully document it well enough that a developer should reasonably be aware of this requirement.

PR #247, opened 2020-02-17
Add ModuleSet object

dadrian:
Eventually this can be used to replace AddCommand, allowing the use of a
non-global config object.

## How to Test

## Notes & Caveats

## Issue Tracking

### DIFF
--- a/module_set.go
+++ b/module_set.go
@@ -1,0 +1,30 @@
+package zgrab2
+
+// ModuleSet is a container holding named scan modules. It is a wrapper around a
+// map.
+type ModuleSet map[string]ScanModule
+
+// AddModule adds m to the ModuleSet, accessible via the given name. If the name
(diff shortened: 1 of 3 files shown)


module_set.go
```diff
@@ -0,0 +1,30 @@
+package zgrab2
+
+// ModuleSet is a container holding named scan modules. It is a wrapper around a
+// map.
+type ModuleSet map[string]ScanModule
```
codyprime: If there is ever going to be a possible writer and reader that access the map concurrently, ModuleSet should be a struct that contains a`sync.Mutex` to guard the map.  e.g.,
```go
type ModuleSet struct {
    lock sync.Mutex // protects Modules
    Modules map[string]ScanModule
}
```
And then of course the methods can all do s.Lock() / s.Unlock() as appropriate.

PR #251, opened 2020-02-18
Remove reference to global config in monitor.go

dadrian:
This updates MakeMonitor() to take the channel size as a parameter,
instead of reading it from the global `config` object. Unfortunately,
the caller of MakeMonitor() doesn't actually have access to the global,
since it's in a different package (bin vs the root package). Luckily,
there doesn't appear to be a reason to have a buffer in this channel.
This updates the caller to pass a hardcoded size of 1.

## Notes & Caveats

@codyprime what do you think about channel size here?

### DIFF
--- a/bin/bin.go
+++ b/bin/bin.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"os"
 	"runtime/pprof"
+	"sync"
 	"time"
 
 	"fmt"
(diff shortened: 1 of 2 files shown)


codyprime:
>@codyprime what do you think about channel size here?

Generally, I think it is best to have as small of a channel size as needed (seems like larger is more likely to mask a sync issue where an empty buffer causes deadlock, etc.), so I like this change.  In this particular case, the `statusesChan` has a writer that is called within a worker goroutine.

I don't see any comments or commit messages in the git history that explains why it is `config.Senders*4`, so I don't know if there was any reasoning for a channel depth 4 times larger than the number of workers (that seems rather specific?). 

I guess one test would be to see if it affects throughput, which I could see being the case if we have workers waiting on the monitor goroutine.

dadrian:
@codyprime I have not tested against a large number of hosts yet. I had the same worry as you, but it should be pretty fast. I think that having a buffer size of 1 effectively forces this to be scheduled before the next scan finishes.

PR #257, opened 2020-03-19
Remove Scanner.Name()

dadrian:
Introduce ScanModuleAttributes

Remove Scanner.Name()

Rather than add a bunch of accessor functions that return static values
for each module, abstract the into ScanModuleAttributes, and add a
single function that exposes this. Previously, much of this information
was exposed as part of the call to RegisterModule. That remains there
for backwards compatibility (for now), but this allows it to be exposed
programatically, and continues work towards removing global state.

Ultimately, name was already tracked separately from the Scanner instance
itself.

## How to Test

Ensure output still works as expected

## Notes & Caveats

This helps with #249, which I have rebased on top of this branch.

### DIFF
--- a/module.go
+++ b/module.go
@@ -11,9 +11,6 @@ type Scanner interface {
 	// subset of the input scan targets
 	InitPerSender(senderID int) error
 
-	// Returns the name passed at init
-	GetName() string
-
 	// Returns the trigger passed at init
 	GetTrigger() string
 
--- a/module.go
+++ b/module.go
@@ -38,6 +38,11 @@ type ScanResponse struct {
 	Error     *string     `json:"error,omitempty"`
 }
 
+type ScanModuleAttributes struct {
+	Name        string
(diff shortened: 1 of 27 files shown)


module.go
```diff
@@ -38,6 +35,11 @@ type ScanResponse struct {
 	Error     *string     `json:"error,omitempty"`
 }
 
+type ScanModuleAttributes struct {
+	Name        string
+	DefaultPort uint
```
codyprime: Is `DefaultPort` intended for use in a future PR to replace the hard-coded default in the modules' calls to `zgrab2.AddCommand`?

codyprime: My only concern is if `DefaultPort` goes into this PR, and subsequent PRs that use it are not merged, it becomes confusing where to change / add the default port for existing / new scanners.  I'd recommend either pulling `DefaultPort` out, or go ahead and use it in the `AddCommand` calls.

PR #262, opened 2020-05-21
Ecubit/fix false positives

elliot-censys:
## Contents
Check contents of banners to make sure they match their protocol before returning success.
Fixes telnet, ssh, smtp, pop3, imap, and dnp3

## How to Test

Test each protocol against a few protocol servers and ssh running on localhost to verify results:
Docker images used:
Telnet: jaredharringtongibbs/telnet-server on :23
pop3, imap: instrumentisto/dovecot on :110 and :143
smtp: bytemark/smtp on :25

Testing for pop3:
$ nc -l 9999 &
$ ./zgrab2 pop3 --port 9999 --timeout 2 <<< "127.0.0.1"
$ ./zgrab2 pop3 --port {23|143|25|22} --timeout 2 <<< "127.0.0.1"
$ > success = false
$ ./zgrab2 pop3 --port 110 --timeout 2 <<< "127.0.0.1"
$ > success = true

Repeat for each protocol, only port running correct protocol should have success=true.

Didn't have a dnp3 server locally to test with; found an open system at 216.164.167.23 to test against for the success case.

### DIFF
--- a/modules/imap/imap.go
+++ b/modules/imap/imap.go
@@ -3,8 +3,6 @@ package imap
 import (
 	"net"
 	"regexp"
-	"strings"
-	"errors"
 	"io"
 
 	"github.com/zmap/zgrab2"
@@ -20,21 +18,6 @@ type Connection struct {
 	Conn net.Conn
 }
 
-// Verify banner begins with a valid IMAP response and handle it
-func VerifyIMAPContents(n int, ret []byte) (string, error) {
-	s := string(ret[:n])
-	if strings.HasPrefix(s, "* OK"){
-		return s, nil
-	}
-	if strings.HasPrefix(s, "* NO"){
-		return s, zgrab2.NewScanError(zgrab2.SCAN_APPLICATION_ERROR, errors.New("IMAP reported error"))
-	}
-	if strings.HasPrefix(s, "* BAD"){
-		return s, zgrab2.NewScanError(zgrab2.SCAN_UNKNOWN_ERROR, errors.New("IMAP request was malformed"))
-	}
-	return s, zgrab2.NewScanError(zgrab2.SCAN_PROTOCOL_ERROR, errors.New("Invalid response for IMAP"))
-}
-
 // ReadResponse reads from the connection until it matches the imapEndRegex. Copied from the original zgrab.
 // TODO: Catch corner cases
 func (conn *Connection) ReadResponse() (string, error) {
@@ -43,7 +26,7 @@ func (conn *Connection) ReadResponse() (string, error) {
 	if err != nil && err != io.EOF && !zgrab2.IsTimeoutError(err) {
 		return "", err
 	}
-	return VerifyIMAPContents(n, ret)
+	return string(ret[:n]), nil
 }
 
 // SendCommand sends a command, followed by a CRLF, then wait for / read the server's response.
--- a/modules/imap/imap.go
+++ b/modules/imap/imap.go
@@ -3,6 +3,9 @@ package imap
 import (
 	"net"
 	"regexp"
+	"strings"
+	"errors"
+	"io"
 
 	"github.com/zmap/zgrab2"
 )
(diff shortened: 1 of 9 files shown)


modules/imap/imap.go
```diff
@@ -17,15 +20,30 @@ type Connection struct {
...
 }
 
+// Verify banner begins with a valid IMAP response and handle it
+func VerifyIMAPContents(n int, ret []byte) (string, error) {
+	s := string(ret[:n])
+	if strings.HasPrefix(s, "* OK"){
+		return s, nil
+	}
+	if strings.HasPrefix(s, "* NO"){
+		return s, zgrab2.NewScanError(zgrab2.SCAN_APPLICATION_ERROR, errors.New("IMAP reported error"))
```
codyprime, line 30: This isn't actually an error, at least for us.  Since the goal of zgrab2 is to identify services, receiving a well-formed error response is a good thing, as it is a positive indication that (in this case) we are talking to an imap server. (This also applies to the `* BAD` case, as well).

There is also `* PREAUTH` and `* BYE`

modules/pop3/pop3.go
```diff
@@ -17,15 +23,28 @@ type Connection struct {
...
 }
 
+// Verifies that a POP3 banner begins with a valid status indicator
+func VerifyPOP3Contents(n int, ret []byte) (string, error) {
+	s := string(ret[0:n])
+	if strings.HasPrefix(s, "+OK "){
+			return s, nil
+	}
+	if strings.HasPrefix(s, "+ERR "){
+			return s, zgrab2.NewScanError(zgrab2.SCAN_APPLICATION_ERROR, errors.New("POP3 Reported Error"))
```
codyprime, line 33: Similar comment as for imap; this is actually a success case

modules/smtp/smtp.go
```diff
@@ -18,15 +20,28 @@ type Connection struct {
...
 
+// Verify that an SMTP code was returned, and that it is a successful one!
+func VerifySMTPContents(n int, ret []byte) (string, error){
+	s := string(ret[:n])
+	code, err := getSMTPCode(s)
+	if err != nil {
+		return s, err
+	}
+	if code < 200 || code >= 300 {
+		return s, zgrab2.NewScanError(zgrab2.SCAN_APPLICATION_ERROR, fmt.Errorf("SMTP returned error code %d", code))
```
codyprime, line 31: Similar comment; if we wanted to vet that it is a valid code, anything returned as a code from the SMTP RFC would qualify as success.  However, I wonder if we even need to be that strict (in case there are non-standard extensions out there)?

modules/pop3/pop3.go
elliot-censys: Based on the contents of the modbus scanner, it seems that SCAN_APPLICATION_ERROR _is_ [a success case](https://github.com/zmap/zgrab2/blob/11425ac78b64e39d95050b042c9505fa28c24641/modules/modbus/scanner.go#L140) - indicating that a valid response was received but the server reported back an error in the final ScanResult

elliot-censys: Though I just realized that string should be "-ERR" regardless

codyprime: `SCAN_APPLICATION_ERROR` can be treated as a "success" error case; the real indication is if the scanner returns `nil` for the results or not.  So it would be OK in this and the IMAP module to return `SCAN_APPLICATION_ERROR` here, except that in both of them `scanner.go` will omit results.

For example, from `pop3/scanner.go`:
```go
    banner, err := conn.ReadResponse()
    if err != nil {
        return zgrab2.TryGetScanStatus(err), nil, err
    }
```

modules/imap/scanner.go
```diff
@@ -148,6 +149,21 @@ func getIMAPError(response string) error {
 	return fmt.Errorf("error: %s", response)
 }
 
+// Check the contents of the IMAP banner and return a relevant ScanStatus
+func VerifyIMAPContents(banner string) zgrab2.ScanStatus {
+	switch {
+	case strings.HasPrefix(banner, "* OK"),
```
codyprime, line 155: Did some searching on censys.io, and some imap servers actually return `OKAY` as well: https://censys.io/ipv4/198.133.159.137/table#143

We should also match on a `strings.Contains(banner, "IMAP)`?
https://censys.io/ipv4/155.230.11.8/table#143

PR #264, opened 2020-05-22
ftp: Add support for implicit TLS

codyprime:
The command line option `--implicit-tls` will attempt connecting via
TLS before intiating the FTP handshake.

## How to Test

You need to find a server that is using TLS around FTP (but not auth-tls or auth-ssl), and run:

`echo ip_address | zgrab2 ftp --port 990 --implicit-tls`

### DIFF
--- a/modules/ftp/scanner.go
+++ b/modules/ftp/scanner.go
@@ -11,6 +11,7 @@
 package ftp
 
 import (
+	"fmt"
 	"net"
 	"regexp"
 	"strings"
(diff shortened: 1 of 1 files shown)


modules/ftp/scanner.go
```diff
@@ -44,8 +49,9 @@ type Flags struct {
 	zgrab2.BaseFlags
 	zgrab2.TLSFlags
 
-	Verbose    bool `long:"verbose" description:"More verbose logging, include debug fields in the scan results"`
-	FTPAuthTLS bool `long:"authtls" description:"Collect FTPS certificates in addition to FTP banners"`
+	Verbose     bool `long:"verbose" description:"More verbose logging, include debug fields in the scan results"`
+	FTPAuthTLS  bool `long:"authtls" description:"Collect FTPS certificates in addition to FTP banners"`
+	ImplicitTLS bool `long:"implicit-tls" description:"Attempt to connect via a TLS wrapped connection first"`
```
bwireman, line 54: `Attempt to connect via a TLS wrapped connection first` implies it will attempt plain text next, which doesn't appear to be the case

codyprime:
rebased to master

PR #275, opened 2020-09-17
Add option to compute http BodySHA256 on decoded BodyText

twschum:
Adds a new way to compute the BodyHash for http scanner.

## How to Test

```
echo <ip> | ./zgrab2 http --port 443 --compute-decoded-body-hash-algorithm=sha256 | jq
 "body_hash": "sha256:ff990fc95970bd2fc6333663ae043aaef308117190507bbd7b4831244cd97c2d",

echo <ip> | ./zgrab2 http --port 443 --compute-decoded-body-hash-algorithm=sha1 | jq
"body_hash": "sha1:a829c753bb3e30296c1faeea6783edf7dd23824f",

./zgrab2 http --port 443 --compute-decoded-body-hash-algorithm=md5 | jq
Invalid value `md5' for option `--compute-decoded-body-hash-algorithm'. Allowed values are: sha256 or sha1
```

## Notes & Caveats

## Issue Tracking

### DIFF
--- a/lib/http/response.go
+++ b/lib/http/response.go
@@ -63,6 +63,8 @@ type Response struct {
 	// BodyHash is the hash digest hex of the decoded http body, formatted `<kind>:<hex>`
 	// e.g. `sha256:deadbeef100020003000400050006000700080009000a000b000c000d000e000`
 	BodyHash string `json:"body_hash,omitempty"`
+	// Number of bytes read from the server and encoded into BodyText
+	BodyTextLength int64 `json:"body_length,omitempty"`
 
 	// ContentLength records the length of the associated content. The
 	// value -1 indicates that the length is unknown. Unless Request.Method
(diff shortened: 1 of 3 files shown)


codyprime:
If we are adding some new options to compute hash, can we also add an option for SHA1 as well?  E.g., a config option like `BodyHashAlgorithm` that can be "sha1" or "sha256", and a corresponding new field in the Response for just `Hash` instead of `BodySHA256`. If the new flag is used populate `Hash` with the specified algorithm, otherwise it behaves as it currently does now.  Also, can the `Hash` field then have a the algorithm prefix (e.g. ."sha1:")

codyprime (approved):

justinbastress (approved):

codyprime (approved):

p-l-:
Quick question: would it be possible to have the possibility to select one or more hash functions, so that we could have MD5, SHA1 **and** SHA256 computed and reported? You never know which hash function are used by other applications.

codyprime:
> Quick question: would it be possible to have the possibility to select one or more hash functions, so that we could have MD5, SHA1 **and** SHA256 computed and reported? You never know which hash function are used by other applications.

I do think that would be nice to have, I guess now as a future enhancement to this patch.

PR #277, opened 2020-10-16
telnet: Optionally return result even on error

codyprime:
Even if the server does not respond to the telnet commands, the
banner may have useful information.  Add the option "--force-banner"
so that the banner string may be returned even on failure.  If the option
is not specified, the origin behavior is unchanged.


bwireman (commented):
Looks like the tests need to updated. LGTM otherwise

codyprime:
> Looks like the tests need to updated. LGTM otherwise

Tests / Travis on zgrab2 have been broken for a while, alas. The Telnet tests themselves actually passed.

bwireman (approved):

PR #282, opened 2020-11-24
Added JARM fingerprinting module

silascutler:
_Added JARM fingerprinting module_

## How to Test
```
echo $(dig +short google.com) | ./zgrab2 jarm
```

## Notes & Caveats
JARM is an active-fingerprinting technique developed by Salesforce.  It was released https://github.com/salesforce/jarm and a write-up on it can be found at: https://engineering.salesforce.com/easily-identify-malicious-servers-on-the-internet-with-jarm-e095edac525a

## Issue Tracking

None

### DIFF
--- a/modules/jarm/scanner.go
+++ b/modules/jarm/scanner.go
@@ -3,51 +3,47 @@
 package jarm
 
 import (
-    "io"
-    "log"
-    "net"
-    _"fmt"
+	_ "fmt"
+	jarm "github.com/RumbleDiscovery/jarm-go"
+	"github.com/zmap/zgrab2"
+	"io"
+	"log"
+	"net"
 	"strings"
-    "github.com/zmap/zgrab2"
-    jarm "github.com/RumbleDiscovery/jarm-go"
-
 )
 
 // Flags give the command-line flags for the banner module.
 type Flags struct {
-    zgrab2.BaseFlags
-    MaxTries int    `long:"max-tries" default:"1" description:"Number of tries for timeouts and connection errors before giving up."`
+	zgrab2.BaseFlags
+	MaxTries int `long:"max-tries" default:"1" description:"Number of tries for timeouts and connection errors before giving up."`
 }
 
 // Module is the implementation of the zgrab2.Module interface.
 type Module struct {
 }
 
-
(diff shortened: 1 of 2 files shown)


modules/jarm/scanner.go
```diff
@@ -0,0 +1,154 @@
...
+                continue
+        }
+
+        ret, err = zgrab2.ReadAvailable(conn)
+        if err != io.EOF && err != nil {
+                rawhashes = append(rawhashes, "")
+                continue
+        }
+
+		ans, err := jarm.ParseServerHello(ret, probe)
```
hdm, line 141: this may need a gofmt pass

silascutler: Good call.  Will update the request

modules/jarm/scanner.go
```diff
@@ -0,0 +1,154 @@
...
+	    }
+	    defer conn.Close()
+
+        _, err = conn.Write([]byte(data))
+        if err != nil {
+                rawhashes = append(rawhashes, "")
+                continue
+        }
+
+        ret, err = zgrab2.ReadAvailable(conn)
```
hdm: this may need multiple reads to match the same buffer size as JARM (1484); others servers with large responses can have different hashes as the truncated response limit is reached

silascutler: What might work would be to instead of using:
```
130 //              ret, err = zgrab2.ReadAvailable(conn)
```
use `ReadAvailableWithOptions()` and specify the buffer size as the
```
130                 ret, err = zgrab2.ReadAvailableWithOptions(conn, 1484, 500*time.Millisecond, 0, 1484)
```


```
func ReadAvailableWithOptions(conn net.Conn, bufferSize int, readTimeout time.Duration, totalTimeout time.Duration, maxReadSize int) ([]byte, error) {
```

silascutler: (So I don't forget - pushed a copy with that change)

silascutler:
bump?

dadrian (approved):

dadrian:
Sorry for the delay on this.

dadrian:
Thanks for the contribution!

PR #286, opened 2020-12-10
Add NativeOS, NTLM, and GroupName to SMBv1 results

elliot-censys:
Performs session setup during SMBv1 negotiations to get these fields.

## How to Test

I tested this locally against [this](https://hub.docker.com/r/dperson/samba) samba image, with the settings modified to allow SMBv1 (it is disabled by default).

via `./zgrab2 smb --setup-session <<< "127.0.0.1,," | jq`:

```
{
  "ip": "x.x.x.x",
  "data": {
    "smb": {
      "status": "success",
      "protocol": "smb",
      "result": {
        "smbv1_support": true,
        "smb_version": {
          "major": 1,
          "minor": 0,
          "revision": 0,
          "version_string": "SMB 1.0"
        },
        "os_name": "Unix",
        "ntlm": "Samba 3.0.37",
        "group_name": "WORKGROUP",
        "has_ntlm": false
      },
      "timestamp": "2020-12-10T16:20:08Z"
    }
  }
}
```

I also tested against various hosts in the wild, though the results aren't interesting.

Against a host that doesn't support SMBv1 or requires credentials during session setup, these string fields will be empty.

## Notes & Caveats

This also includes a small bugfix; in the SMBv1 negotiation, a `ByteCount` field was set to 14 rather than the correct 12.

### DIFF
--- a/lib/smb/smb/smb.go
+++ b/lib/smb/smb/smb.go
@@ -95,6 +95,10 @@ const (
 	ShareCapAsymmetric             uint32 = 0x00000080
 )
 
+const (
+	SmbHeaderV1Length = 32
+)
+
 type HeaderV1 struct {
 	ProtocolID       []byte `smb:"fixed:4"`
 	Command          uint8
@@ -133,6 +137,24 @@ type NegotiateReqV1 struct {
 	Dialects  []uint8 `smb:"fixed:12"`
 }
 
+type SessionSetupV1Req struct {
+	HeaderV1
+	WordCount             uint8
+	AndCommand            uint8
+	Reserved1             uint8
+	AndOffset             uint16
+	MaxBuffer             uint16
+	MaxMPXCount           uint16
+	VCNumber              uint16
+	SessionKey            uint32
+	OEMPasswordLength     uint16
+	UnicodePasswordLength uint16
+	Reserved2             uint32
+	Capabilities          uint32
+	ByteCount             uint16
+	VarData               []byte
+}
+
 type NegotiateResV1 struct {
 	HeaderV1
 	WordCount       uint8
@@ -147,8 +169,9 @@ type NegotiateResV1 struct {
 	SystemTime      uint64
 	ServerTimezon   uint16
 	ChallengeLength uint8
-	ByteCount       uint16
-	// variable data afterwords that we don't care about
+	ByteCount       uint16 `smb:"len:VarData"`
+	// variable data afterwards that we don't care about
+	VarData []byte
 }
 
 type NegotiateReq struct {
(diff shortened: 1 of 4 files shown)


lib/smb/smb/smb.go
```diff
@@ -147,8 +169,9 @@ type NegotiateResV1 struct {
 	SystemTime      uint64
 	ServerTimezon   uint16
 	ChallengeLength uint8
-	ByteCount       uint16
-	// variable data afterwords that we don't care about
+	ByteCount       uint16 `smb:"len:VarData"`
+	// variable data afterwards that we don't care about
```
bwireman, line 173: We can probably remove this comment

bwireman:
I think you need to update the zschema as well. @justinbastress or @codyprime correct me if I'm wrong

codyprime:
Protocol changes look good, but `zgrab2_schemas/zgrab2/smb.py` needs to be updated with the new fields.

lib/smb/smb/zgrab.go
```diff
@@ -269,6 +279,62 @@ func (ls *LoggedSession) LoggedNegotiateProtocolv1(setup bool) error {
...
+	// When using unicode, a padding byte will exist after the header
+	paddingLength := int((buf[11] >> 7) & 1)
+	// Skip header
+	buf = buf[SmbHeaderV1Length:]
+	// The byte after the header holds the number of words in uint16s
+	if len(buf) < (int(buf[0])*2)+3+paddingLength {
+		return nil
+	}
+	// words + 3 bytes for wordlength & bytecount + potential unicode padding
+	buf = buf[(int(buf[0])*2)+3+paddingLength:]
```
codyprime, line 307: This index value (`(int(buf[0])*2)+3+paddingLength:`) is validated above, but I recommend moving this out to a variable.  That will make it easier to verify that the validation is happening against the same value.

elliot-censys: Good call

zgrab2_schemas/zgrab2/smb.py
```diff
@@ -62,7 +65,7 @@ def extended(base, new):
             "smb_encryption_support": Boolean(doc="Server supports encryption"),
             }, doc="Capabilities flags for the connection. See [MS-SMB2] Sect. 2.2.4."),
         'negotiation_log': negotiate_log,
-        'has_ntlm': Boolean(),
+        'has_ntlm': Boolean(doc="Server supports the NTLM authentication method"),
```
elliot-censys, line 68: There _will_ be cases where `ntlm` is populated but `has_ntlm` is false, so some clarification is needed

codyprime (approved):

PR #292, opened 2021-01-05
Add checkpoint banner grabbing

Te-k:
Checkpoint has a specific admin protocol that provides the hostname of the firewall, here is a PR adding a module to zgrab to scan and get this hostname. It is based on several tools like [this metasploit module](https://github.com/rapid7/metasploit-framework/blob/master/modules/auxiliary/gather/checkpoint_hostname.rb)

## How to Test

`./zgrab2 checkpoint`

## Notes & Caveats

* I am not very familiar with go, so this definitely need some code review
* I have not implemented any integration test as I don't think there is any open source implementation of this protocol. It would be easy to write a python script to reproduce the protocol but then you are testing what you know works, so I don't think it would be very useful
* I am not fully sure the output schema is correct, I have not tested it.

So this PR needs a bit more work from the community before being merged.

### DIFF
--- a/modules/checkpoint.go
+++ b/modules/checkpoint.go
@@ -1,0 +1,7 @@
+package modules
+
+import "github.com/zmap/zgrab2/modules/checkpoint"
+
+func init() {
+	checkpoint.RegisterModule()
+}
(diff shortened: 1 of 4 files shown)


phillip-stephens:
Apologies for taking so long to get to this, I've added integration tests and everything based off your initial commit, gave acknowledgements in my PR. Closing this since #728 supersedes

PR #293, opened 2021-01-12
Update links in "Output schema" section of README.md

themaxdavitt:
Update some dead links in the README.

## How to Test

Compare the links. :)

## Notes & Caveats

Just found out about the ZMap project and was reading through some of the repos, so I'm assuming these are the correct new links. I'm sorry if not!

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,9 +123,9 @@ func init() {
 
 ### Output schema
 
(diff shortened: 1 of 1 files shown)


themaxdavitt:
Actually, it looks like a ton of the links in the README.md file in that directory are broken, which I don't feel knowledgeable enough to fix without introducing new issues.

PR #295, opened 2021-02-02
improves http body decoding and enforces max length

bwireman:
improves http body decoding and enforces max length

## How to Test
```
echo "www.qq.com" | ./zgrab2 http -p 443 --use-https --max-size=64 --with-body-size | jq .
```

## Notes & Caveats

I decided the initial read should still be limited and that we should just re-enforce length afterwards, but I can see the argument for not doing that

qq.com is a good example because it decodes correctly even though it isn't certain

### DIFF
--- a/modules/http/scanner.go
+++ b/modules/http/scanner.go
@@ -425,25 +425,36 @@ func (scan *scan) Grab() *zgrab2.ScanError {
 	if resp.ContentLength >= 0 && resp.ContentLength < maxReadLen {
 		readLen = resp.ContentLength
 	}
-	// EOF ignored here because that's the way it was, CopyN goes up to readLen bytes
-	bytesRead, _ := io.CopyN(buf, resp.Body, readLen)
-	if scan.scanner.config.WithBodyLength {
-		scan.results.Response.BodyTextLength = bytesRead
-	}
-	bufAsString := buf.String()
+	io.CopyN(buf, resp.Body, readLen)
+	encoder, encoding, certain := charset.DetermineEncoding(buf.Bytes(), resp.Header.Get("content-type"))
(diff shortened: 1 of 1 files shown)


modules/http/scanner.go
```diff
@@ -425,25 +425,36 @@ func (scan *scan) Grab() *zgrab2.ScanError {
...
 
-	// do best effort attempt to determine the response's encoding
-	// ignore the certainty and just go with it
-	encoder, _, _ := charset.DetermineEncoding(buf.Bytes(), resp.Header.Get("content_type"))
+	bodyText := ""
+	decodedSuccessfully := false
 	decoder := encoder.NewDecoder()
+	
+	//"windows-1252" is the default value and will likely not decode correctly
+	if certain || encoding != "windows-1252" {
```
codyprime, line 436: Why won't `windows-1252` decode correctly?  Did you run into specific issues with that encoding?

bwireman: Sorry I should have explained this better. `windows-1252` is what determineEncoding returns when it has no idea and is causing issues like the one we saw this morning. So unless it's certain I don't want to use it

bwireman: https://github.com/golang/net/blob/4acb7895a057/html/charset/charset.go#L103

codyprime: An edge case would be a detection of `windows-1252` earlier than the default (via `prescan`, I guess).  But given the returns of charset, I guess there is no other way to know that this is the default rather than computed guess (other than manually re-implementing charset.DetermineEncoding).  So your solution is probably best.

codyprime (approved):

PR #297, opened 2021-02-04
fix for smtp module

aspacewalz:
The SMTP module was matching on "STMP" when verifying the contents of the scan response. This PR fixes the typo and adds a test for the VerifySMTPContents() function.

## How to Test

`make test` and verify all tests pass.


codyprime (approved):
Looks good, and thanks for creating a test for it.

PR #299, opened 2021-02-09
Fix SMB scanner marshaller panic

elliot-censys:
The SMB scanner's encoder would panic when unmarshalling dynamic length content that was assigned a length larger than the actual buffer.

## How to Test

This should not pass (before the change, it would panic):

`go test lib/smb/ntlmssp`

## Notes & Caveats

With this change alone, we will still falsely say an SMB host with this challenge is not SMB, but the scanner will no longer panic, which is nice.

The cause of the panic was an invalid `AvPair` with very large length:

```
 id       length   ?????
 vvvvvvv  vvvvvvv  vvvvvvvvv
"\x17\xe9\x0e\x91\x31\xe7\xb2\xce\xac\x29\x59\xba"
```

which is not documented by Microsoft. The other `AvPair`s here are valid.

Options to properly solve this:

1. Ignore these failures on `AvPair` unmarshalling since we don't report/need the values from `AvPairs` anyways.

Positive: 
- We will mark this host as SMB
- We will still require the rest of `Challenge` required to be unmarshalled correctly

Negative:
- There may be false positives as a result; unsure how crucial this field is to identifying SMB

2. Find what that opcode is and why it is valid (as well as others that aren't documented)

Positive:
- We will mark this host as SMB
- This `Challenge` field will be populated for weird hosts like this

Negative:
- This might be a lot of work
- There isn't a cap on "what opcodes might be out there?"

Naively, it seems there won't be many false positives as a result of ignoring that field's errors, but I'm happy to be corrected; these cases also seem pretty rare, so I'm not sure how much it matters.


codyprime:
I think the approach you've implemented is an OK trade-off, given that this is not hit too often

codyprime (approved):

Issue #300, opened 2021-02-24
SNI server name not updated on redirects causing handshake failure


Rich5:
Steps to reproduce:

`echo 148.62.23.244,granite-logistics.com | ./zgrab2 http --max-redirects 10 --retry-https
`

Results:

```
INFO[0000] started grab at 2021-02-24T00:23:24Z
{
  "ip": "148.62.23.244",
  "domain": "granite-logistics.com",
  "data": {
    "http": {
      "status": "unknown-error",
      "protocol": "http",
      "result": {},
      "timestamp": "2021-02-24T00:23:24Z",
      "error": "tls: oversized record received with length 20527"
    }
  }
}
INFO[0006] finished grab at 2021-02-24T00:23:30Z
{"statuses":{"http":{"successes":0,"failures":1}},"start":"2021-02-24T00:23:24Z","end":"2021-02-24T00:23:30Z","duration":"6.184948073s"}
```

I see references to this error in various places in other projects so it makes me think it may be in some deep library, but any idea why this is happening? The only other reference in zgrab2 is here [issue](https://github.com/zmap/zgrab2/issues/94) from 2018.

Rich5:
As an update if I remove --retry-http there are a series of redirects ultimately leading to cloudflare CDN and end with the error 

"error": "remote error: handshake failure"

But still doesn't explain why retrying with https give the tls oversized error.

mzpqnxow:
@Rich5 thanks for sending this in, you actually found a really nice bug that may be having a big impact across large datasets. The workaround for now is to use `--no-sni`. If you want to read in more detail feel free but that's all you need to know. If you don't mind, can you rename this issue something like "SNI server name not updated on redirects causing handshake failure", so it's clear by the title what the root cause is?

There are a few things going on here. It seems that SNI is the root cause of everything. The bug (specifically) is that  zgrab2 is using SNI by default- but in the case of redirects from one SSL endpoint to another, the SNI server name value does not get updated after the first redirects- so for servers that are using SNI, you'll either get the wrong certificate or (in this case) the handshake will fail completely

The invalid length issue is expected behavior, but I understand why it's confusing you, or seems wrong, if you're curious

For your case, the 301/302 chain goes:

http://granite-logistics.com:80 -> https://granite-logistics.com:443 -> https://www.granite-logistics.com:443

The handshake is failing on the last one because it's sending an SNI server name of "granite-logistics.com" when it *should* be sending an SNI server name of "www.granite-logistics.com"

[...]

PR #306, opened 2021-03-22
[BUGFIX] Set SNI on each redirect to avoid handshake failures / incorrect server name

mzpqnxow:
This is a fix for issue #300, where the SNI server name was only set once, at the start of an HTTP scan. This caused incorrect SNI server name values when not using `--no-sni` or `--server-name`. The example case is in #300

Details follow, the initial commit was a botched patch but the latest commit fixes it. It was a little more involved than I thought it would be, though still a pretty small change


mzpqnxow:
This is a bad fix... I'm working on doing it correct, as is it will send the addr value (which is host *and* port) as the SNI name, which will cause the server to fallback to a default certificate in this case

mzpqnxow:
@rich5 my first patch was bad. This version should fully fix your issues. Can you please test it when you have a moment? It's working for me on your sample hosts and does not break `--no-sni` or `--server-name` if you're depending on those for other uses

@dadrian, @elliot-censys, @zakird I think this patch is important as it may impact a non-negligible amount of HTTPS endpoints. Would you take a look at it if you have time? It's a small patch- I included comments that you can remove if you like. The only thing that a mistake in this patch could impact is SNI-related functionality- specifically `--no-sni` and `--server-name`. I tested both and they seem to be working as expected

I needed to make a minor change to function parameters for the function returned by `getTLSDialer()` because there was a namespace collision with the `net` module and a parameter- I just renamed the parameter from `net` to `network`

The gist (behaviorally) can be seen using the example @rich5 provided in #300:

```
echo granite-logistics.com | zgrab2 http --max-redirects 10`
```

For this endpoint, the behavior before the patch looked like this:

```
http://granite-logistics.com 302 -> https://granite-logistics.com (SNI=granite-logistics.com) 302 --> https://www.granite-logistics.com (SNI=granite-logistics.com) : SNI FAIL
```

This patch just updates the SNI server name on every request in a redirect chain (unless `--no-sni` or `--server-name` are set) so that in the example, the finaly request to https://www.granite-logistics.com has the correct SNI server name value set, allowing the negotiation to complete. This can be confirmed by checking that the correct content was grabbed and/or by checking the protocol negotiation on the wire to see the SNI server name value

The important point here is that the old behavior causes servers that have mandatory SNI to either fail negotiation entirely or return a fallback certificate and fallback virtual host configuration

This specific case happened to be Cloudflare, which obviously makes up a big chunk of HTTPS on the web and is only getting bigger. I believe Cloudflare customers need to explicitly opt-in to SNI- they *definitely* need to opt-in to an SNI-only configuration- so it may not be too much of a problem now, but it's going to get bigger over time as they are pushing SNI pretty hard. They'are also pushing ESNI but I believe this will just be available opportunistically alongside SNI because hardly any browsers or client libraries support it.

While we're here- I want to point out two other behaviors I noticed, both with and without this change

1. Use of `--server-name` sets the SNI server name on *all* HTTPS connections, not just the initial one. This may be intentional / well-known, but it caught me off guard and I think in general it's not how most users would want it to behave. That said, use of `--server-name` is probably not very widespread so it's not a big deal. I'm happy to open it as an issue if there's any agreement that it's something that should be changed
2. A failure anywhere in the redirect chain causes the `--retry-https` logic to fire, even if the first request elicited a successful plaintext HTTP response. I think it would make sense to either change the default behavior to not retry as https in this case, or maybe add a flag (somewhat similar to `--redirects-succeed`) to prevent it. It seems to be that in 99% of cases this is just wasted bandwidth and/or a very confusing error message when the SSL handshake tries to interpret `P/` as an SSL message length. I opened up #307 to get an opinion on that behavior

Thanks, I appreciate your patience and help with everything- I can imagine that features/bugfixes aside, just "keeping the lights on" for zcrypto, zmap and zgrab2 probably takes up a lot of your time

BTW- If anyone would like to take a look at #302 that would be great too. It's a much simpler change :)

engn33r:
I encountered no issues when testing this PR in my use cases :+1:

modules/http/scanner.go
```diff
@@ -239,18 +239,33 @@ func (scan *scan) dialContext(ctx context.Context, network string, addr string)
...
+		//  - host is the current target of the request in this context; this is true for the
+		//    initial request as well as subsequent requests caused by redirects
+		//  - scan.scanner.config.ServerName is the value from --server-name if one was specified
+
+		// If SNI is enabled and --server-name is not set, use the target host for the SNI server name
+		if !scan.scanner.config.NoSNI && scan.scanner.config.ServerName == "" {
+			host, _, err := net.SplitHostPort(addr)
+			if err != nil {
+				log.Errorf("http/scanner.go getTLSDialer: unable to split host:port '%s'", addr)
+				log.Errorf("Something went wrong splitting host/port: %s", err)
```
dadrian, line 264: Can you condense this down into one log message?

mzpqnxow: Absolutely, meant to remove that, nice catch. Should be fixed now

mzpqnxow: BTW, I think I also need to add logic to check if `host` is a an IP address. Currently, this will send an SNI server name value of the dotted quad (if a dotted quad is what is being used for the `Host:` header)

From [RFC4366](https://tools.ietf.org/html/rfc4366):

> Literal IPv4 and IPv6 addresses are not permitted in "HostName".

I think I need to add a check to see if it's an IP address and if it is, not set anything at all. Based on a quick test with this build and looking at a wire capture, I confirmed that with a build of this PR it is in fact setting the server name to a dotted quad when an IP only is specified

I think the answer is obviously, just checking in case you have some specific knowledge about how SNI works today and whether a deviation from the RFC would be a normal thing

dadrian (approved):

Issue #273, opened 2020-08-25
[PR Open] Improving retry-https by reading HTTP 400 and/or adding `--retry-http`


mzpqnxow:
The `retry-https` feature of the `zgrab2` `http` module is very useful when surveying a large collection of non-standard port. Based on measurements that I do regularly, it seems HTTPS (as opposed to HTTP) on non-standard ports is becoming more and more common, where it used to be almost entirely HTTP

However, the current implementation of `retry-https` only retries as HTTPS when the plaintext request does not result in an RFC compliant HTTP response. By design, it does not look at the response code or message to make any decisions. This was a reasonable choice to me but I am now encountering quite a bit of cases where additional/different logic could make `zgrab2` more effective in dealing with this

A specific case of the problem this approach runs into is the below. This is from nginx in a pretty basic/default HTTPS configuration. I'm not sure about how common it is with Apache or other servers. I'm happy to provide some numbers from a large sample set to give an idea of the magnitude of the problem, or on what other server software exhibits this behavior

The following shows a plaintext HTTP request made to an HTTPS endpoint. NGINX returns an RFC compliant HTTP response back- specifically, an HTTP 400 with a clear description of what happened- HTTP request made to SSL endpoint

```
> GET / HTTP/1.1
> Host: somesite.com:8443
> User-Agent: curl/7.64.0
> Accept: */*
> 
< HTTP/1.1 400 Bad Request
< Server: nginx
< Date: Tue, 25 Aug 2020 13:46:39 GMT
< Content-Type: text/html
< Content-Length: 248
< Connection: close
< 
<html>
<head><title>400 The plain HTTP request was sent to HTTPS port</title></head>
<body>
<center><h1>400 Bad Request</h1></center>
<center>The plain HTTP request was sent to HTTPS port</center>
<hr><center>nginx</center>
</body>
</html>
```

Per [RFC7231, 6.5.1](https://tools.ietf.org/html/rfc7231#section-6.5.1):

> The 400 (Bad Request) status code indicates that the server cannot or
> will not process the request due to something that is perceived to be
> a client error (e.g., malformed request syntax, invalid request
> message framing, or deceptive request routing).

So clearly HTTP 400 does not necessarily mean an HTTP request was sent to an HTTPS listener, but it may be the most common meaning on the modern Internet for HTTP 400 statistically. I'm not sure but could gather some statistics

[...]

PR #308, opened 2021-03-29
Feature fail http to https for known status code and substrings

mzpqnxow:
This implements what is described in #273. If zgrab2 http was invoked with `--retry-https`, the response code to a request is 400 *AND*  one of the three known substrings is present, consider it as a protocol failure to cause a retry as HTTPS

## How to Test
Use against an nginx or Apache server that speaks HTTPS. Specify `--retry-https` and the flag implemented here (`--fail-http-to-https`). Rather than storing the 400 response that suggests you retry the request as HTTPS, zgrab2 will automatically retry as HTTPS

## Notes & Caveats

This is not a generic solution- it's a specific workaround for specific HTTP servers- nginx and Apache. I believe that the substrings included have very good coverage for this sort of behavior based on some sample data I collected

I considered what impact this might have on performance, but I figured since the function this happens in is responsible for performing the hashing of the body the limited logic and string comparisons shouldn't be much overhead. To be on the safe side, I slice the string down to a smaller size since the sub-strings it looks for are always towards the beginning of the response

## Issue Tracking

Issue #273


modules/http/scanner.go
```diff
@@ -446,6 +448,23 @@ func (scan *scan) Grab() *zgrab2.ScanError {
...
 
+	// Application-specific logic for retrying HTTP as HTTPS; if condition matches, return protocol error
+	if scan.scanner.config.FailHTTPToHTTPS && scan.results.Response.StatusCode == 400 && readLen < 1024 && readLen > 24 {
+		// Apache: "You're speaking plain HTTP to an SSL-enabled server port"
+		// NGINX: "The plain HTTP request was sent to HTTPS port"
+		var sliceLen int64 = 128
+		if readLen < sliceLen {
+			sliceLen = readLen
+		}
+		sliceBuf := bodyText[0:sliceLen]
```
vl4deee11, line 459: you can replace with this `sliceBuf := bodyText[:sliceLen]`

mzpqnxow: Thanks @vl4deee11, just committed. I'm very slowly picking up go as you can tell :>

mzpqnxow:
@dadrian what are your thoughts on this? Is it a reasonable feature to include for the http module? I've gotten a good deal of use out of it and couldn't find any alternative aside from postprocessing results and running against and forcing https for the servers that behaved this way

mzpqnxow:
poke :)

zakird (approved):

Issue #303, opened 2021-03-13
Replace TravisCI with Github Actions


dadrian:
TravisCI just doesn't work nowadays. Nearly every run times out.

phillip-stephens:
Looks like we've moved all tests to GitHub actions now and they gate the PR's. Closing this issue.

PR #309, opened 2021-04-14
First draft of GitHub Action

engn33r:
Aim to address #303 by migrating from TravisCI to GitHub Actions, because many issues with TravisCI [have recently been reported](https://news.ycombinator.com/item?id=25338983).

## How to Test

While in PR form, the GitHub Action build results will not be visible, since this repo is currently using only TravisCI. You will need to test this PR on your own branch and push the updated branch to GitHub to see the results in the Actions tab of GitHub.

After this action is part of the repo, it will be triggered by submitting a new PR or commit. The results will be seen in the Actions tab of this repo.

## Notes & Caveats

**Warning**: This PR does not have full CI functionality in its initial form. The tests are currently encountering errors. I am no expert with Go build systems, so `continue-on-error: true` is used for the two test steps to get a passing result while we troubleshoot the errors encountered. I will probably need assistance from others to resolve these errors. If someone wants to borrow this code and create a new PR that works, go ahead. The sooner we have CI properly working again, the better.

## Issue Tracking

Issue #303

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,9 +10,6 @@ jobs:
         with:
           go-version: 1.16
 
-      - name: Set up Docker for integration tests
-        uses: docker-practice/actions-setup-docker@master
-
       - name: Check out source
         uses: actions/checkout@v2
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,36 @@
+name: ZGrab2 build action
+on: [push, pull_request]
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.16
+
+      - name: Set up Docker for integration tests
+        uses: docker-practice/actions-setup-docker@master
+
+      - name: Check out source
+        uses: actions/checkout@v2
+
+      - name: Install goimports
+        run: sudo apt install golang-golang-x-tools
+
+      - name: Build
+        run: |
+          go get -t ./...
+          make
+
+      - name: Test
+        run: make clean zgrab2 test
+        continue-on-error: true
+
+      - name: Integration test
+        run: make integration-test
+        continue-on-error: true
+
+      - name: Integration cleanup
+        run: ./integration_tests/cleanup.sh
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,21 +16,7 @@ jobs:
       - name: Check out source
         uses: actions/checkout@v2
 
-      - name: Install goimports
-        run: sudo apt install golang-golang-x-tools
-
       - name: Build
         run: |
           go get -t ./...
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,36 @@
...
+      - name: Test
+        run: make clean zgrab2 test
+        continue-on-error: true
+
+      - name: Integration test
+        run: make integration-test
+        continue-on-error: true
+
+      - name: Integration cleanup
+        run: ./integration_tests/cleanup.sh
```
dadrian, line 36: I think we can just drop the integration tests from CI. I'm pretty sure we broke them when we switched to Go modules. Did you have any luck getting them to run locally?

.github/workflows/build.yml
```diff
@@ -0,0 +1,36 @@
...
+          go-version: 1.16
+
+      - name: Set up Docker for integration tests
+        uses: docker-practice/actions-setup-docker@master
+
+      - name: Check out source
+        uses: actions/checkout@v2
+
+      - name: Install goimports
+        run: sudo apt install golang-golang-x-tools
```
dadrian, line 20: Why do we need `goimports`?

dadrian (commented):
Thanks for doing this! Comments inline.

.github/workflows/build.yml
engn33r: No, I didn't successfully get this running locally. If you or someone else has a better understanding of how it should work, that might help with troubleshooting. I thought about reverting the repo to the last commit where TravisCI successfully passed and checking if the tests worked locally for me then, but I didn't go that far.

Since I did not get these tests working, I could remove my edits of the runner-base.Dockerfile file. I only modified it while trying to get it working and it's possible my edits, while minor, might move the file further from a functioning state.

.github/workflows/build.yml
engn33r: Apparently goimports didn't need to be installed separately in TravisCI (based on the .travis.yml), but it does in GitHub Actions. It is used in the Makefile: https://github.com/zmap/zgrab2/blob/5e9507cacf0a7f5cd57aa17a48b29ce6cde76de9/Makefile#L21

I didn't try to understand why the Makefile was the way it was, so it's possible that the Makefile needs cleaning up as well.

.github/workflows/build.yml
dadrian: I'd leave the Dockerfile alone, and drop the integration tests from CI. Can you file a follow up issue to figure out:

1. What they did
2. How to run them now that Go modules exist?

@justinbastress might have insights, he wrote them originally, back in the dark days of Go development.

.github/workflows/build.yml
dadrian: Let's skip checking formatting in CI for now.

engn33r:
I minified the changes that this PR introduces and created a new issue around the tests I couldn't get working. Let me know if I missed anything.

dadrian (approved):

PR #314, opened 2021-06-01
Add error handling for FTP TLS handshake

hmcguinn:
This pull requests adds a check in modules/ftp/scanner.go for an error upon completing a TLS handshake.

### DIFF
--- a/modules/ftp/scanner.go
+++ b/modules/ftp/scanner.go
@@ -265,7 +265,10 @@ func (s *Scanner) Scan(t zgrab2.ScanTarget) (status zgrab2.ScanStatus, result in
 		}
 		results.ImplicitTLS = true
 		results.TLSLog = tlsConn.GetLog()
-		tlsConn.Handshake()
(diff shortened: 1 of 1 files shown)


zakird:
This looks like it will prevent attempting to grab banner if the TLS handshake fails. I believe that this is the right thing to do since it's a TLS wrapped FTP session not running TLS after FTP has started. @codyprime can you confirm that this makes sense?

zakird (approved):
LGTM unless @codyprime flags something in the logic.

codyprime (approved):
looks good to me as well, thanks

codyprime:
> This looks like it will prevent attempting to grab banner if the TLS handshake fails. I believe that this is the right thing to do since it's a TLS wrapped FTP session not running TLS after FTP has started. @codyprime can you confirm that this makes sense?

Yeah, I also believe this is the correct thing to do, since in this path TLS was requested for this scan.  We should error out if we cannot establish a TLS connection.

PR #328, opened 2021-09-30
jarm: update jarm to not fail on handshake failure

aspacewalz:
This PR removes the error handling for a JARM probe response. The reason is that a handshake failure should still be processed the same as any other response. Currently a handshake failure results in the ZeroHash value being returned. This does not match the salesforce jarm module which should be treated as the correct implementation.

After this change, the salesforce and zgrab2 modules will both return the same fingerprint for services that return a handshake failure to one of the probes.

## How to Test

go test ./...

## Notes & Caveats

This PR is being created to bring the zgrab2 JARM module in line with the behavior of the salesforce jarm tool.

### DIFF
--- a/modules/jarm/scanner.go
+++ b/modules/jarm/scanner.go
@@ -6,7 +6,6 @@ import (
 	_ "fmt"
 	jarm "github.com/RumbleDiscovery/jarm-go"
 	"github.com/zmap/zgrab2"
-	"io"
 	"log"
 	"net"
 	"strings"
@@ -103,13 +102,8 @@ func (scanner *Scanner) Scan(target zgrab2.ScanTarget) (zgrab2.ScanStatus, inter
 	// Stores raw hashes returned from parsing each protocols Hello message
 	rawhashes := []string{}
 
-	// Stores final module results
-	r := Results{}
-
 	// Loop through each Probe type
 	for _, probe := range jarm.GetProbes(target.Host(), int(scanner.GetPort())) {
-		data := jarm.BuildProbe(probe)
-
 		var (
 			conn net.Conn
 			err  error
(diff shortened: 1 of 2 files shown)


modules/jarm/scanner.go
```diff
@@ -120,20 +114,14 @@ func (scanner *Scanner) Scan(target zgrab2.ScanTarget) (zgrab2.ScanStatus, inter
...
 		}
 
-		// ret, err = zgrab2.ReadAvailable(conn)
-		ret, err = zgrab2.ReadAvailableWithOptions(conn, 1484, 500*time.Millisecond, 0, 1484)
-		if err != io.EOF && err != nil {
-			rawhashes = append(rawhashes, "")
-			conn.Close()
-			continue
-		}
+		ret, _ = zgrab2.ReadAvailableWithOptions(conn, 1484, 500*time.Millisecond, 0, 1484)
```
aspacewalz, line 124: It appears to be the case that error responses from the target should still be processed by `ParseServerHello` which contains logic for handling the various failure responses appropriately.

modules/jarm/scanner.go
```diff
@@ -142,12 +130,11 @@ func (scanner *Scanner) Scan(target zgrab2.ScanTarget) (zgrab2.ScanStatus, inter
 			continue
 		}
 
-		rawhashes = append(rawhashes, string(ans))
+		rawhashes = append(rawhashes, ans)
```
aspacewalz, line 133: `ans` is already a string. no need to cast.

modules/jarm/scanner.go
```diff
@@ -120,20 +114,14 @@ func (scanner *Scanner) Scan(target zgrab2.ScanTarget) (zgrab2.ScanStatus, inter
 			return zgrab2.TryGetScanStatus(err), nil, err
 		}
 
-		_, err = conn.Write([]byte(data))
+		_, err = conn.Write(jarm.BuildProbe(probe))
```
aspacewalz, line 117: data was already a byte slice and was also only used in this function call.

utility.go
```diff
@@ -145,9 +145,7 @@ func ReadAvailableWithOptions(conn net.Conn, bufferSize int, readTimeout time.Du
 			}
 			return ret, err
 		}
-		if err != nil {
-			return ret, err
-		}
+
```
aspacewalz, line 148: In this block `err` is **ALWAYS** `nil` so i removed it.

codyprime (approved):
Code cleanup changes look good.

Tested against a known problem host, verified the changes after matched the salesforce jarm tool fingerprint.

Issue #252, opened 2020-02-25
Zgrab2 + TLS CLient Cert Auth causes failure, but curl and openssl s_client succeed 


mzpqnxow:
# Zgrab2 + TLS CLient Cert Auth Causes Failure, But curl and openssl s_client Succeed

[This bit about golang and session renegotiation shines some light on things](https://gist.github.com/vbatts/0afaee5c394b52a95f2d)
[This might be helpful enough for me to produce a PR](https://github.com/golang/go/commit/af125a5193c75dd59307fcf1b26d885010ce8bfd)

So I found a bit of a strange edge-case with how zgrab2 handles TLS client certificate authentication failure in certain server/load-balancer configurations. In short, `curl`, `openssl s_client -connect`, etc. retrieve HTML content but `zgrab2` does not; it returns an empty response (see below)

The example site is _https://cn-stg-mbr-migration-websvc-scus.azurewebsites.net_
The error emitted from zgrab2 is `local error: no renegotiation`. More detail is below

## Curl Behavior - Successfully retrieves HTML

Perform a standard HTTP 1.1 GET request on the server. It successfully retrieves HTML, though it contains an error message.

```
$ curl --http1.1 -k https://cn-stg-mbr-migration-websvc-scus.azurewebsites.net 2>/dev/null
<!DOCTYPE html>
<html>
<head>
    <title>Web App - Unavailable</title>
...
<body bgcolor="#00abec">
    <div id="feature">
        <div id="content">
            <h1 id="unavailable">Error 403 - Forbidden: Client Certificate Required.</h1>
            <p id="tryAgain">The web app you have attempted to reach requires a client certificate and does not accept any requests without one.</p>
        </div>
    </div>
</body>
</html>
$ 
```
## Zgrab2 Behavior - Fails to retreive HTML

When using zgrab2 to try to retrieve this HTML content, it fails. No response data is set and the error is described as `local error: no renegotiation`

```
› echo cn-stg-mbr-migration-websvc-scus.azurewebsites.net | zgrab2 http -p 443 --use-https --max-redirects=5 2>/dev/null
{"domain":"cn-stg-mbr-migration-websvc-scus.azurewebsites.net","data":{"http":{"status":"unknown-error","protocol":"http","result":{},"timestamp":"2020-02-25T13:19:51-05:00","error":"local error: no renegotiation"}}}
```

[...]

PR #334, opened 2021-11-16
Implement --negotiate-freely to address server renegotiation requests / "no renegotiation" error

mzpqnxow:
Feature as requested in #252, which is required for some endpoints. Without this, negotiation fails with:

```
"error":"local error: no renegotiation"
```

## How to Test

To reproduce a failure, you'll need to find an endpoint that has the behavior of requesting a renegotiation during the handshake. I have an example here that I found by grepping for "tls: no renegotiation" in a sample dataset I had handy:

```
$ echo mx-stg-mbr-migration-websvc-scus.azurewebsites.net  | ./zgrab2 http --use-https -p 443                                                               
INFO[0000] started grab at 2021-11-16T10:24:12-05:00    
{"domain":"mx-stg-mbr-migration-websvc-scus.azurewebsites.net","data":{"http":{"status":"unknown-error","protocol":"http","result":{},"timestamp":"2021-11-16T10:24:12-05:00","error":"local error: tls: no renegotiation"}}}
```

With the patch applied, and with the `--renegotiate-freely` flag, it works as expected:

```
$ echo mx-stg-mbr-migration-websvc-scus.azurewebsites.net  | ./zgrab2 http --use-https -p 443 --renegotiate-freely
INFO[0000] started grab at 2021-11-16T10:24:54-05:00    
{"domain":"mx-stg-mbr-migration-websvc-scus.azurewebsites.net","data":{"http":{"status":"success","protocol":"http","result":{"response":{"status_line":"403 Forbidden","status_code":403,"protocol":{"name":"HTTP/1.1","major":1,"minor":1},"headers":{"content_length":["1976"],"content_type":["text/html"],"date":["Tue, 16 Nov 2021 15:24:54 GMT"],"server":
...
```

## Notes & Caveats

I think it may be best to remove the flag and make this a default setting rather than a flag. There may be potential security concerns by allowing unlimited renegotiation requests, but the goal of zgrab2 (and zcrypto) is to be as compatible as possible, not to negotiate unbreakable TLS sessions, obviously :)

In addition, I'll add that there are three settings for this behavior in golang tls- disallow, allow once, or allow freely- as many times as the server requests. I think it's best to either allow it as many times as the server wants or not at all

## Issue Tracking

#252


mzpqnxow:
I have some high level data on how common this configuration is and where it tends to exist, for those interested

Based on the input I have, most of the sites with this renegotiation behavior are in Azure

1. Azure Traffic Manager (trafficmanager.net)
2. Azure Cloud App / Cloud Gateway (azuregateway-*.cloudapp.net)
3. Azure Websites (azurewebsites.net)

I see a very small handful of hosts that do *not* have Azure names or are not in Azure,, but the majority are

It's a relatively small number of hosts in my scope, only about a half a percent across ~75k endpoints

mzpqnxow:
@dadrian @zakird any interest in meeting this one? The implementation is straightforward, I just need to fix the merge conflicts that have popped up since I sent it

Thanks!

mzpqnxow:
@dadrian @zakird sorry to be persistent, just a nudge for this one

More than any of the other things I've sent along, I think this feature is critical in for comprehensive coverage with the `zgrab2` HTTPS captures. Configurations with this behavior may not comprise the *majority* of HTTPS services out there, but they're non-zero in my scope of assets, which adds up at scale

Thanks as always for your patience and consideration with PRs :)

dadrian:
Sorry for the delay. Do feel free to @ me on general, it's more likely I'll notice. This seems like a reasonable idea. I can do a review once merge cleanup is fixed, please @ me again.

mzpqnxow:
> Sorry for the delay. Do feel free to @ me on general, it's more likely I'll notice. This seems like a reasonable idea. I can do a review once merge cleanup is fixed, please @ me again.

Terrific, thanks! I generally hesitate to poke people but I will keep this in mind for issues like this

mzpqnxow:
I'm closing this but will reenter a new one, I got all mixed up with the changes to the heartbleed option ...

PR #336, opened 2021-11-29
Add TLS logs in results when using the banner module with the `--tls` flag

marc-etienne:
This patch adds the result of the TLS handshake when using the banner module with the `--tls` option.

## How to Test

`echo 4.5.6.7 | zgrab2 banner --tls -p 443` shows the certificates of the TLS connection if it succeeded.

## Notes & Caveats

Hopefully that won't break modules that could be subclassing the Banner module, but the changes are minimal to avoid breaking anything.

I based the patch on existing code that was done for SMTP with STARTTLS, HTTPS, etc.

Something to note: results are returned even if no data is received in the underlying TLS connection. This means that if the banner module is used without `--tls` and no data is received `data.banner.result` in the JSON will be an empty object (`{}`) instead of being omitted. I'm not sure if that's a problem or not, but we'd need to add some additional `if` statements if result must not be `{}`.

## Issue Tracking

I don't think there is an issue regarding this.

### DIFF
--- a/modules/banner/scanner.go
+++ b/modules/banner/scanner.go
@@ -44,6 +44,8 @@ type Scanner struct {
 type Results struct {
 	Banner string `json:"banner,omitempty"`
 	Length int    `json:"length,omitempty"`
+
+	TLSLog *zgrab2.TLSLog `json:"tls,omitempty"`
 }
 
 // RegisterModule is called by modules/banner.go to register the scanner.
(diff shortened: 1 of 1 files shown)


zakird (approved):

Seanstoppable:
I believe an implementation of this already has been merged

phillip-stephens:
Yep, looks like this was added in PR #404. Apologies we didn't get this merged earlier since this was opened in 2021 😅.

marc-etienne:
😂 No worries, important thing is that everyone gets TLS logs now 🎉

you_get_and_everyone_gets

PR #350, opened 2022-04-28
ipp: nameLength should be uint16, not int16

codyprime:
nameLength is used for slice allocation.  If the value is unmarshaled
as a negative value, then a runtime error will occur:

   runtime error: makeslice: len out of range

(this commit also includes some minor formatting cleanup)

## How to Test

I have been unable to reproduce the incident I encountered again, but you can attempt scanning an IPP service found on Censys to verify there has not been a regression: 
https://search.censys.io/search?resource=hosts&sort=RELEVANCE&per_page=25&virtual_hosts=EXCLUDE&q=services.service_name%3A+IPP

### DIFF
--- a/modules/ipp/scanner.go
+++ b/modules/ipp/scanner.go
@@ -41,8 +41,18 @@ var (
 	// TODO: Explain this error
 	ErrVersionNotSupported = errors.New("IPP version not supported")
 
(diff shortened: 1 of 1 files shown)


elliotcubit:
Looks good to me, formatting changes only whitespace and this should be `unsigned`

elliotcubit (approved):

PR #354, opened 2022-06-22
Add docker support

creekorful:
This PR allows zgrab2 to be built as a docker image so that it can be run like this for example:

```
$ echo 'xxx.xxx.xxx.xxx' | docker run -i zgrab2:latest ssh 2>/dev/null | jq
```
it could be nice to add a GitHub action that automagically build and push the docker image every time a tag is created, so that people may enjoy an official image :)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+Dockerfile
+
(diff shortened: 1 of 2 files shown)


creekorful:
Eager for feedback !

dadrian (approved):

Issue #318, opened 2021-06-23
SSH: Unable to marshal data


TrueSkrillor:
Hey everyone,

I stumbled across a fatal error while performing an internet-wide ssh scan as part of my masters thesis resulting in premature end of the scan. After some time, zgrab2 logs the following error and quits:

`FATA[10636] unable to marshal data: json: error calling MarshalJSON for type ssh.kexAlgorithm: json: error calling MarshalJSON for type ssh.PublicKey: json: error calling MarshalJSON for type time.Time: Time.MarshalJSON: year outside of range [0,9999]`

As the error message indicates the timestamp of the public key seems to be invalid (not within the valid range). By extending the call to the logging function I was able to isolate the list of IPs that caused this issue. As it turned out only six IPs in the entire IPv4  address space caused this issue but I will try to avoid sharing the IPs publicy due to obvious reasons.
For debugging purposes I performed a SSH handshake using OpenSSH with verbose output enabled (which was possible). See below for the captured output (I stripped the log file a little but it should still contain all necessary information):

```
OpenSSH_8.6p1, OpenSSL 1.1.1k 25 Mar 2021
debug1: Connecting to x.x.x.x [x.x.x.x] port 22.
debug1: Connection established.
debug1: Local version string SSH-2.0-OpenSSH_8.6
debug1: Remote protocol version 2.0, remote software version Go
debug1: compat_banner: no match: Go
debug1: Authenticating to x.x.x.x as 'none'
debug1: SSH2_MSG_KEXINIT sent
debug1: SSH2_MSG_KEXINIT received
debug1: kex: algorithm: curve25519-sha256@libssh.org
debug1: kex: host key algorithm: ecdsa-sha2-nistp384-cert-v01@openssh.com
debug1: kex: server->client cipher: chacha20-poly1305@openssh.com MAC: <implicit> compression: none
debug1: kex: client->server cipher: chacha20-poly1305@openssh.com MAC: <implicit> compression: none
debug1: expecting SSH2_MSG_KEX_ECDH_REPLY
debug1: SSH2_MSG_KEX_ECDH_REPLY received
debug1: Server host certificate: ecdsa-sha2-nistp384-cert-v01@openssh.com SHA256:[...], serial [...] ID "[...]" CA ssh-ed25519 SHA256:[...] valid from 2021-01-21T00:01:13 to 2038-01-19T04:14:07
debug1: No matching CA found. Retry with plain key
```

[...]

PR #355, opened 2022-06-22
Make it a non-fatal error when EncodeGrab fails to marshal data

jamadden:
Make it a non-fatal error when `EncodeGrab` fails to marshal data when called by `grabTarget`. Previously, if a result couldn't be encoded, the process crashed. Now, it logs an error and continues.

The proximate cause for this is https://github.com/zmap/zgrab2/issues/318 (SSH certificates with dates too large crash the process), and the workaround is due to @TrueSkrillor in that same issue. The source of the error is the go standard library's [time module](https://cs.opensource.google/go/go/+/master:src/time/time.go;drc=41b9d8c75e45636a153c2a31d117196a22a7fc6c;l=1317), and it points us to https://github.com/golang/go/issues/4556#c15. Essentially, time is trying to produce a RFC3339 string, and refuses to do so with a year that's out of range.

We began seeing this problem about the week of June 6, 2022, and this change got us past that problem.

## How to Test

The most direct test would be to scan a server vending certificates that trigger this problem. That's potentially hard because I can't provide a list of such servers (they weren't consistent from one day to the next in our testing). Otherwise, I'm not sure how to get the invalid data into `grabTarget`; I admit I haven't looked at the test suite for this project at all (sorry!).

## Notes & Caveats

However, this does appear to be a global change, applying to all scanners. We aren't currently affected by this (or we would have had other uses of zgrab2 crashing with a similar error), but in the future, should we hit bad data, we won't crash, we'll just get a logged warning that we may not notice. That's both good and bad, and I certainly understand if this doesn't get merged because of its global nature.

I don't know go well. I was a bit concerned about passing what must be the `nil` return value back from `grabTarget`, into the `outputQueue` `byte[]` channel, and ultimately into the `OutputResults` function which wants to write it to the (buffered) output file. However, from reading the go stdlib's source, it appears that, because `len(v)`, where v is a `byte[]`, is 0, nothing actually happens: the empty result is swallowed by the buffered writer.

## Issue Tracking

https://github.com/zmap/zgrab2/issues/318

### DIFF
--- a/processing.go
+++ b/processing.go
@@ -178,7 +178,7 @@ func grabTarget(input ScanTarget, m *Monitor) []byte {
 	raw := BuildGrabFromInputResponse(&input, moduleResult)
 	result, err := EncodeGrab(raw, includeDebugOutput())
 	if err != nil {
-		log.Fatalf("unable to marshal data: %s", err)
+		log.Errorf("unable to marshal data: %s", err)
 	}
 
 	return result


dadrian (approved):

jamadden:
Thanks!

Issue #351, opened 2022-05-09
HTTP module and --source-ip= not working on multi-homed system


mzpqnxow:
I have a system with two WAN links. I noticed that `--source-ip=` does not have any effect in this configuration. I confirmed this using the `http` module by looking at the traffic leaving the respective interfaces (they're separate physical NICs, not aliased)

I have not yet looked at the cause of this but will give it a shot when I have time. If anyone else can take a look that'd be swell :)

Thanks

dadrian:
Source IP does not work at all and should be removed due to limitations with Go. If you need this behavior, use Linux network namespaces instead.

mzpqnxow:
> Source IP does not work at all and should be removed due to limitations with Go. If you need this behavior, use Linux network namespaces instead.

Thanks for this, I hadn't thought of it but it's quite a nice solution and (as I'm sure you know) can be done in a very lightweight manner without any dependencies, using only `ip netns` and network namespaces (no need for "container" solutions, nor filesystem namespaces, user/process namespaces, etc.)

For the benefit of others who may need to do this, here's the idiot's guide to doing it quickly in the form of a script

The caveat (if it isn't obvious) is that the interface will no longer be available in the primary namespace- so don't do it on your management/primary NIC by accident...

```bash
#!/bin/bash
set -euo pipefail
declare -r IFACE=eth0
declare -r NS=zg_ns
declare -r IP=1.2.3.4/20
declare -r GW=1.2.3.1
declare -r SUDO=sudo
declare -r SCAN_USER=user
# Create new network namespace (delete if already exists)
$SUDO ip netns del $NS &>/dev/null || true
$SUDO ip netns add $NS
# "move" the interface from the primary namespace into the new namespace
$SUDO ip link set $IFACE netns $NS
# Configure the interface as usual ...
$SUDO ip netns exec $NS ip addr add $IP dev $IFACE
$SUDO ip netns exec $NS ip link set $IFACE up
$SUDO ip netns exec $NS ip route add default via $GW dev $IFACE
# Enter the namespace and test networking (or run zgrab2, or an interactive shell)
echo Testing WAN connectivity inside $NS ...
$SUDO ip netns exec $NS sudo -u $SCAN_USER ping -c 3 8.8.8.8
# $SUDO ip netns exec $NS sudo -u $SCAN_USER -i
read -p "Press enter to destroy the namespace, control-C to leave it.."
$SUDO ip netns del $NS
[...]

PR #359, opened 2022-07-02
Feature/source interface (inferior solution for #351, *if* unprivileged users can't create namespaces)

mzpqnxow:
This is a "feature" which is useful but only in very specific configurations which are described as a comment in the code. I'm not insistent on this being merged, please feel free to close it and let it just live in the closed issues as a reference

This stems from #351 and was implemented before a better solution was suggested by @dadrian

Merging it probably does no significant harm but because it's also rarely helpful and is a somewhat half-baked solution, it's perfectly reasonable to not want to accept it

I'll as another point against accepting it that it is not a very clean implementation- it uses an environment variable instead of a command-line flag, the error checking is questionable, and assumptions are made (like using the first address for an interface) that may cause it to not work well with aliased interfaces (e.g. eth0:1, eth0:2)

This has only been tested on my configuration, where it worked well enough to unblock me. I didn't implement it as a flag both because I am lazy and because (similar to the original `--source-ip` flag) it is difficult to describe it in in the usage function without it confusing users who almost certainly don't want what it provides

From the comments:

```golang
	// There are caveats here and it's designed for a very specific use-case
	// Use-case is for machines with multiple default routes. Consider a system
	// with mgmt0 and wan0
	//   - Both are technically WAN links with public IP addresses
	//   - Both have an entry in the routing table for 0/0
	//   - Only one (wan0) is intended for network probes
	//   - It is not an option to simply reorder the routes system-wide, for
	//     whatever reason
	//
	// Assuming the above all hold, you can use $ZIFACE to override the
	// interface that the TCP/IP stack would choose by default (which is based
        // entirely on the priority of the routing table)
	// 
	// This is a fix ("hack", really) for the issue described in:
	//   https://github.com/zmap/zgrab2/issues/351
	//
	// @dadrien suggested a much less invasive and more flexible solution, which
	// is spinning up a network namespace with only the desired interface and
	// using that for the scan. That solves this problem without any code changes.
	//
	// That works very well and is by far the best solution *except* where it is not
	// available, in cases where the system is hardened via:
	//
	//   kernel.unprivileged_userns_clone = 0
	//
	// In this case, you can fall back on this hacky mechanism, by setting the
	// $ZIFACE environment variable to the interface you want to prefer over the
	// others
```

Thanks!


conn.go
```diff
@@ -342,6 +344,53 @@ func (d *Dialer) SetDefaults() *Dialer {
...
+	if iface := os.Getenv("ZIFACE"); iface != "" {
+	    ief, err := net.InterfaceByName(iface)
+	    if err != nil {
+			panic("InterfaceByName(): interface specified does not exist")
+	    }
+		addrs, err := ief.Addrs()
+		if err != nil || len(addrs) == 0 {
+			panic("Addrs(): specified interface does not exist or has addresses assigned")
+		}
+		d.Dialer.LocalAddr = addrs[0]
```
dadrian, line 385: So my worry is that this dialer won't end up being used in all cases (specifically, any scanner that makes multiple connections)

dadrian (changes requested):

mzpqnxow:
It’s probably best to close this, the suggestion @dadrian offered requires no code changes, let alone ones as hacky as this one

thanks again @dadrian for the netns idea!

PR #383, opened 2023-07-11
Fix build instructions

Weltraumschaf:
The documented way with `go get` gives an error message:

> go get github.com/zmap/zgrab2
> go: go.mod file not found in current directory or any parent directory.
>        'go get' is no longer supported outside a module.
>        To build and install a command, use 'go install' with a version,
>        like 'go install example.com/cmd@latest'
>        For more information, see https://golang.org/doc/go-get-install-deprecation
>        or run 'go help get' or 'go help install'.

I'm not very experienced with go. Worked works for me: Clone repo and run `make`.


dadrian (approved):

Weltraumschaf:
Hi, what's up?

developStorm:
Sorry for the delay. Thanks for helping us with the documentation!

Issue #344, opened 2022-03-12
Add port field to input format


burpheart:
zgrab2 cannot simply import the results of other scanners.
Each ip may have a large number of different ports and not just one or a few fixed ports.

mzpqnxow:
> zgrab2 cannot simply import the results of other scanners.
> 
> Each ip may have a large number of different ports and not just one or a few fixed ports. 

This is already supported and can be done using the CSV-style target file with triggers/tags. Each trigger should have a corresponding ini file specifying the module and port for that trigger. You can then invoke 'zgrab2' with the 'multi' module, and specify the targets file and respective ini file

See #221 for an example of how I do this. The easiest way to do it (while avoiding issues 'zgrab2' has when loading extremely large target and ini files) is to have one ini file per-port. You then specify the trigger in the target file (the trigger will effectively represent the port for that host)

I feel that triggers and the CSV-style target format is under-utilized, but absolutely critical for certain usage patterns. As an example (my use-case) when looking at large target sets, with varying ports, where some ports benefit from specific settings. For example, for port 80, you may want 'retry-https', while for port 443, you may want 'use-ssl'

This is a bit more work than what you may have had in mind, but it's a one-time effort to write the script

While I originally thought the solution you described would be best and simple to implement (basically, allowing the user to specify <host>:<port> for each target, rather than <host>) I learned after working a bit on the code that it is fundamentally incompatible with that approach. It's not an oversight, it was a design decision from what I can tell, and it has both good and bad tradeoffs

Hope this is helpful

mzpqnxow:
I think this can be closed

PR #397, opened 2023-10-12
Add port number to CSV input and output

AnthraX1:
This PR allows setting a custom port in the CSV input and it also shows the port in output results.
The port number is added to the last field for backward compatibility reasons. However it's ugly when you need just IP, Port. You have to input IP,,,Port instead. You can take the target port number out in the JSON output of zgrab2 later.

## How to Test

Domain and port:
echo ',google.com,,80' | ./zgrab2 http

IP and port:
echo '1.2.3.4,,,80' |  ./zgrab2 http

## Notes & Caveats

Port number won't be shown in the output if the input was not in CSV format. For example, when using just IP or domain per line and using -p option to supply the port number.

## Issue Tracking
[https://github.com/zmap/zgrab2/issues/344](https://github.com/zmap/zgrab2/issues/344)

### DIFF
--- a/input.go
+++ b/input.go
@@ -5,29 +5,31 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"strconv"
 	"strings"
 
 	log "github.com/sirupsen/logrus"
 )
 
 // ParseCSVTarget takes a record from a CSV-format input file and
-// returns the specified ipnet, domain, and tag, or an error.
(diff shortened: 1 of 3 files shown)


AnthraX1:
I'm closing this one as it causes null pointer deref in some modules.

zakird (approved):

Issue #378, opened 2023-04-08
Can we disable tls parsing in http module


rony0724:
something like --no-tls 

`./zgrab http -p 443 --use-https --no-tls`

At some ips, tls certificate parsing is getting failed. 

{"http":{"status":"unknown-error","protocol":"http","result":{},"timestamp":"2023-04-08T08:53:52Z","error":"tls: failed to parse certificate from server: asn1: syntax error: PrintableString contains invalid character"}

mzpqnxow:
Can you provide enough info to reproduce this? Is there a host/port with this behavior you can share?

I would be much more interested in fixing it rather than masking it

rony0724:
You can try the command, I think it is specifically because of tls certificate is not following the asn.1 standard. 
So we have a option to disable the tis parsing we can avoid this in Golang.

echo 34.107.18.135 | ./zgrab http -p 443 --use-https

dadrian:
The fix for this would be to loosen the ASN1 parsing in https://github.com/zmap/zcrypto

mzpqnxow:
I believe this can now be resolved

mzpqnxow:
@rony0724 have you tried using since the changes referenced here were merged? They should provide the behavior you were looking for

PR #400, opened 2023-11-28
Permissive TLS parsing

Seanstoppable:
This is driven by  #378, https://github.com/zmap/zcrypto/issues/364 and https://github.com/zmap/zgrab2/pull/334

This allows a number of scans to actually succeed, rather than fail out when parsing the certificate

Example without permissive parsing:

```
echo FAILING_IP | ./zgrab2 http -p 443 --use-https
INFO[0000] started grab at 2023-09-21T21:25:29-05:00
{"ip":"FAILING_IP","data":{"http":{"status":"unknown-error","protocol":"http","result":{},"timestamp":"2023-09-21T21:25:29-05:00","error":"tls: failed to parse certificate from server: asn1: structure error: explicitly tagged member didn't match"}}}
INFO[0001] finished grab at 2023-09-21T21:25:29-05:00
{"statuses":{"http":{"successes":0,"failures":1}},"start":"2023-09-21T21:25:29-05:00","end":"2023-09-21T21:25:29-05:00","duration":"987.606886ms"}
```

With Permissive parsing:

```
echo FAILING_IP | ./zgrab2 http -p 443 --use-https
INFO[0000] started grab at 2023-09-21T21:25:34-05:00
{"ip":"FAILING_UP","data":{"http":{"status":"application-error","protocol":"http","result":{"response":{"status_line":"302 Found","status_code":302,"protocol":{"name":"HTTP/1.1","major":1,"minor":1},"headers":{"content_length":["0"],
... all the HTTP and TLS handshake log data
```

### DIFF
--- a/tls.go
+++ b/tls.go
@@ -12,8 +12,10 @@ import (
 	"time"
 
 	log "github.com/sirupsen/logrus"
+	"github.com/zmap/zcrypto/encoding/asn1"
 	"github.com/zmap/zcrypto/tls"
 	"github.com/zmap/zcrypto/x509"
+	"github.com/zmap/zcrypto/x509/pkix"
 )
 
 // Shared code for TLS scans.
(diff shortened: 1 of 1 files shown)


dadrian:
We could probably default to this tbh

dadrian (approved):

mzpqnxow:
> We could probably default to this tbh

That would be awesome- I wasn't sure how it might impact the output schemas that y'all have, and if it might break any test-cases you have, so I didn't want to be presumptuous

Probably the vast majority of users would be appreciative if it as a default

Seanstoppable:
I'm happy to change this to be default behavior if that is the desire. Just opted to do behavioral backwards compatibility.

Seanstoppable:
Updated to just be the default

PR #411, opened 2024-02-18
Add test steps to build

Seanstoppable:
Adds test running to the CI step

A number of recent commits have broken tests
It would be nice if this didn't happen and the build told us.


Seanstoppable:
Note that this currently fails due to changes introduced in #375, which I also highlight in #410

developStorm (approved):

PR #416, opened 2024-04-03
fix: close CPU profile

testwill:
_Add a description of your changes here._

## How to Test

## Notes & Caveats

## Issue Tracking


developStorm:
Thanks for the PR! 🚀 Just a quick note: `pprof.StartCPUProfile()` is non-blocking, so deferring the file close right after might cut off data logging prematurely. Maybe we can close the file in stopCPUProfile() instead? That way, we ensure everything is logged properly before closing. Thoughts?

PR #418, opened 2024-04-11
refactor: improve and align ftp tls error msg

### LINKED ISSUE #276: ftp module: status success when AUTH TLS failed
I noticed an inconsistency among the smtp, imap, pop3 and ftp modules: If explicit tls is selected, the smtp, imap and pop3 modules will not set .data.PROTO.status to "success" if the explicit TLS handshake fails. But the ftp module will happily do so. You have to filter for key_material or server_finished in the tls handshake log to get the actually successful connections. Just something odd I noticed, and wanted to let you know in case you want to keep it consistent (which I would suggest). This is with version 0.1.1, which I realize is quite old by now, but I am in the middle of something and can't update right now, sorry.

Thanks!

developStorm:
resolves #276

### DIFF
--- a/modules/ftp/scanner.go
+++ b/modules/ftp/scanner.go
@@ -216,14 +216,14 @@ func (ftp *Connection) GetFTPSCertificates() error {
 	ftpsReady, err := ftp.SetupFTPS()
 
 	if err != nil {
-		return err
+		return fmt.Errorf("error setting up FTPS: %w", err)
 	}
 	if !ftpsReady {
 		return nil
 	}
 	var conn *zgrab2.TLSConnection
 	if conn, err = ftp.config.TLSFlags.GetTLSConnection(ftp.conn); err != nil {
-		return err
+		return fmt.Errorf("error setting up TLS connection: %w", err)
 	}
 	ftp.results.TLSLog = conn.GetLog()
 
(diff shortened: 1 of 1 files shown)


modules/ftp/scanner.go
```diff
@@ -279,7 +279,7 @@ func (s *Scanner) Scan(t zgrab2.ScanTarget) (status zgrab2.ScanStatus, result in
 	}
 	if s.config.FTPAuthTLS && is200Banner {
 		if err := ftp.GetFTPSCertificates(); err != nil {
-			return zgrab2.SCAN_APPLICATION_ERROR, &ftp.results, err
+			return zgrab2.TryGetScanStatus(err), &ftp.results, err
```
phillip-stephens, line 282: Let's wrap the errors returned in error checking blocks throughout the `Scan` function so we can disambiguate where the error originated during debugging.

phillip-stephens (changes requested):
While we're here, let's wrap the errors in the function `Scan()` as well. It looks like [here](https://github.com/zmap/zgrab2/blob/master/scanner.go#L39-L40) we're returning the error string, so it'd be helpful for debugging if we knew exactly where the error came from.

phillip-stephens (approved):
LGTM, thanks for the improvement! @developStorm

PR #425, opened 2024-04-24
patch: getVersionNumber function

developStorm:
to construct version number with major, minor, and patch from each individual bytes in Ausbg fields

as suggested in https://github.com/zmap/zgrab2/pull/423#issuecomment-2071678271


phillip-stephens (approved):
LGTM, thanks @developStorm

PR #437, opened 2024-05-21
style: reformat all go files

developStorm:
And activate go formatter workflow


phillip-stephens (changes requested):
Overall looks good, thanks for taking this up @developStorm !

One thing, could we add a `make lint` similar to [here](https://github.com/zmap/zdns/pull/375/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR58) so it's easier to do this locally without hunting for the `gofmt` command?

developStorm:
Thanks for reviewing @phillip-stephens! Implemented in 18d5069a990b51f5e53bf0f59c627e076a089e4f. I did only go fmt since linter is out of scope of this PR...

phillip-stephens (approved):
Ran `gofmt` myself and confirmed the change-set is the same. LGTM

PR #461, opened 2024-08-19
Try to align golang around a single version

Seanstoppable:
Currently, we have version defined in:
go.mod/go.sum (1.12)
.github actions (1.18)
Dockerfile (1.20)

This PR tries to align them to the latest version, which is the Dockerfile, this should help with some of the library updates by dependabot, as at least one of them (prometheus) fails due to requiring golang 1.20
I try to have github actions read the golang version, to reduce maintenance

I updated go.mod to update 1.12 to 1.20, and then ran `go mod tidy` which added everything else.

_Add a description of your changes here._

## How to Test

## Notes & Caveats

## Issue Tracking


Seanstoppable:
Note that a LOT of extra versions of libraries seem to be dropped. These appear to largely be dropped if we "only" align to the github pipelines version of 1.18, which also happens to resolve a number of CVEs my company's code scanner is picking up like:
https://github.com/advisories/GHSA-c3h9-896r-86jm
https://github.com/advisories/GHSA-9phm-fm57-rhg8
As these versions/libraries are removed

phillip-stephens (approved):
LGTM, thanks for this!!!

Issue #452, opened 2024-06-11
ZGrab performance issue with http?


phillip-stephens:
Not a bug per se, but was investigating a slow zgrab scan and noticed that it maxes out at sending 6 MB/s. While there might be a lot of CPU overhead, this just seems egregiously slow and surely there's some performance gains we can achieve.

<img width="1044" alt="image" src="https://github.com/zmap/zgrab2/assets/23459798/9b251979-3f9f-4f12-94d1-0deb12409513">

CLI command - `cat ~/100k-domains.txt | ./zgrab2 http --max-redirects=3`

mzpqnxow:
> Not a bug per se, but was investigating a slow zgrab scan and noticed that it maxes out at sending 6 MB/s. While there might be a lot of CPU overhead, this just seems egregiously slow and surely there's some performance gains we can achieve. This issue is to track that.
> 
> 
> 
> <img width="1044" alt="image" src="https://github.com/zmap/zgrab2/assets/23459798/9b251979-3f9f-4f12-94d1-0deb12409513">
> 
> 
> 
> CLI command - `cat ~/100k-domains.txt | ./zgrab2 http --max-redirects=3`
> 
> 
> 
> This was run on a VM with a 1 Gb/s + link and plenty of cores(22)/RAM (56 GB)
> 
> 

Does using the "senders" global (-s if I remember correctly..) not help with this?

Regardless, it seems like a rather unusual number as a default... unless it's as a ratio to the number of logical cores. If that's the case, it does in fact seem to be very low, assuming your measurement is correct

I will re-read what you said, but is this worker's performing DNS lookups or the actual HTTP requests? I'm asking because in the past, users have created issues that were ultimately caused by DNS lookups

I've managed to avoid DNS issues by pre-resolving targets, then using the targets file format with both the Virtual Host/FQDN and IP address, so DNS is only needed on redirects. This is necessary anyway if you want to cover all of the IP addresses for load-balanced FQDNs that return more than one IP in the A record lookup

Seanstoppable:
I recently noticed that we do DNS lookups whenever we get a redirect, irrespective of if we are going to follow it, which is probably at least part of the problem.

Seanstoppable:
I've tracked this to this line:
https://github.com/zmap/zgrab2/pull/462/files#diff-dddc8e7eb09019004b9ea48a345b3fb794995c7b661e5e2aa0494d56d6ce7bd3L542
[...]

PR #462, opened 2024-08-19
Fixes a few issues with redirects

Seanstoppable:
First, this prevents a DNS lookup from happening when we encounter a redirect, *even if we don't intend to follow it*. This likely addresses some part of #452

Second, if we aren't following redirects, don't have the scan fail in an 'application-error'. We are succeeding in what we intended to do, which is to scan without following redirects

## How to Test

I tested by scanning targets that do redirects both with and without this change, using wireshark to monitor for DNS lookups and reviewing output to ensure that we are the same (within reason for randomized items like cookies and dates)


Seanstoppable:
Actually, I apparently miss-tested. I think some DNS caching is invalidating my wireshark manual validation, and DNS lookups are still being done somewhere in the process.

phillip-stephens (approved):
LGTM

PR #477, opened 2024-12-31
gofmt -w -r 'interface{} -> any'

thecsw:
Change all occurrences of `interface{}` to `any` in the codebase. These are the same.

See that Russ Cox already changed all the occurrences of `interface{}` in golang source code -> https://github.com/golang/go/commit/2580d0e08d5e9f979b943758d3c49877fb2324cb

There really is no need to use `interface{}` anymore, since we have a simpler alias provided for it by default.

Efforts to convert stragglers are continuing, see for example changes by Ian Lance Taylor in https://github.com/golang/go/commit/2580d0e08d5e9f979b943758d3c49877fb2324cb

## How to Test

N/A

## Notes & Caveats

N/A

## Issue Tracking

N/A

### DIFF
--- a/bin/bin.go
+++ b/bin/bin.go
@@ -114,7 +114,7 @@ func ZGrab2Main() {
 	if m, ok := flag.(*zgrab2.MultipleCommand); ok {
 		iniParser := zgrab2.NewIniParser()
 		var modTypes []string
-		var flagsReturned []interface{}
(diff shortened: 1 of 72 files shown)


thecsw:
Closing because now a duplicate of https://github.com/zmap/zgrab2/pull/479.

PR #479, opened 2025-01-16
Group Basic/TLS Flag Groupings in Modules and interface{} -> any

phillip-stephens:
- Replace all instances of `interface{}` with the new (as of Go 1.18, not too new) `any` for readability
- Add Flag `group` tags to group Basic/TLS Flags together. Makes it easier for CLI users to find the commands they're looking for

Example of new `--help`
```zsh
$ make; ./zgrab2 redis --help             
...
Usage:
  zgrab2 [OPTIONS] redis [redis-OPTIONS]

Probe for Redis

Application Options:
  -o, --output-file=                   Output filename, use - for stdout (default: -)
  -f, --input-file=                    Input filename, use - for stdin (default: -)
  -m, --metadata-file=                 Metadata filename, use - for stderr (default: -)
  -l, --log-file=                      Log filename, use - for stderr (default: -)
  -s, --senders=                       Number of send goroutines to use (default: 1000)
      --debug                          Include debug fields in the output.
      --flush                          Flush after each line of output.
      --gomaxprocs=                    Set GOMAXPROCS (default: 0)
      --connections-per-host=          Number of times to connect to each host (results in more output) (default: 1)
      --read-limit-per-host=           Maximum total kilobytes to read for a single host (default 96kb) (default: 96)
      --prometheus=                    Address to use for Prometheus server (e.g. localhost:8080). If empty, Prometheus is disabled.
      --dns=                           Address of a custom DNS server for lookups. Default port is 53.

Help Options:
  -h, --help                           Show this help message

[redis command options]
          --custom-commands=           Pathname for JSON/YAML file that contains extra commands to execute. WARNING: This is sent in the clear.
          --mappings=                  Pathname for JSON/YAML file that contains mappings for command names.
          --max-input-file-size=       Maximum size for either input file. (default: 102400)
          --password=                  Set a password to use to authenticate to the server. WARNING: This is sent in the clear.
          --inline                     Send commands using the inline syntax
          --verbose                    More verbose logging, include debug fields in the scan results
          --use-tls                    Sends probe with a TLS connection. Loads TLS module command options.

    Basic Options:
      -p, --port=                      Specify port to grab on (default: 6379)
      -m, --maxbytes=                  Maximum byte read limit per scan (0 = defaults)
      -n, --name=                      Specify name for output json, only necessary if scanning multiple modules (default: redis)
      -t, --timeout=                   Set connection timeout (0 = no timeout) (default: 10s)
      -g, --trigger=                   Invoke only on targets with specified tag

    TLS Options:
          --heartbleed                 Check if server is vulnerable to Heartbleed
          --session-ticket             Send support for TLS Session Tickets and output ticket if presented
          --extended-master-secret     Offer RFC 7627 Extended Master Secret extension
          --extended-random            Send TLS Extended Random Extension
          --no-sni                     Do not send domain name in TLS Handshake regardless of whether known
          --sct                        Request Signed Certificate Timestamps during TLS Handshake
          --keep-client-logs           Include the client-side logs in the TLS handshake
          --time=                      Explicit request time to use, instead of clock. YYYYMMDDhhmmss format.
          --certificates=              Set of certificates to present to the server
          --certificate-map=           A file mapping server names to certificates
          --root-cas=                  Set of certificates to use when verifying server certificates
          --next-protos=               A list of supported application-level protocols
          --server-name=               Server name used for certificate verification and (optionally) SNI
          --verify-server-certificate  If set, the scan will fail if the server certificate does not match the server-name, or does not chain to a trusted root.
          --cipher-suite=              A comma-delimited list of hex cipher suites to advertise.
          --min-version=               The minimum SSL/TLS version that is acceptable. 0 means that SSLv3 is the minimum.
          --max-version=               The maximum SSL/TLS version that is acceptable. 0 means use the highest supported value.
          --curve-preferences=         A list of elliptic curves used in an ECDHE handshake, in order of preference.
          --no-ecdhe                   Do not allow ECDHE handshakes
          --signature-algorithms=      Signature and hash algorithms that are acceptable
          --heartbeat-enabled          If set, include the heartbeat extension
          --dsa-enabled                Accept server DSA keys
          --client-random=             Set an explicit Client Random (base64 encoded)
          --client-hello=              Set an explicit ClientHello (base64 encoded)

```

### DIFF
--- a/bin/bin.go
+++ b/bin/bin.go
@@ -114,7 +114,7 @@ func ZGrab2Main() {
 	if m, ok := flag.(*zgrab2.MultipleCommand); ok {
 		iniParser := zgrab2.NewIniParser()
 		var modTypes []string
-		var flagsReturned []interface{}
+		var flagsReturned []any
 		if m.ConfigFileName == "-" {
 			modTypes, flagsReturned, err = iniParser.Parse(os.Stdin)
 		} else {
--- a/conn_bytelimit_test.go
+++ b/conn_bytelimit_test.go
@@ -65,7 +65,7 @@ type readLimitTestConfig struct {
 // Call sendReceive(), and check that the input/output match, and that any expected errors / truncation occurs.
 func checkedSendReceive(t *testing.T, conn *TimeoutConnection, size int) (result error) {
 	// helper to report + return an error
-	tErrorf := func(format string, args ...interface{}) error {
+	tErrorf := func(format string, args ...any) error {
 		result = fmt.Errorf(format, args)
 		t.Error(result)
 		return result
--- a/integration_tests/.template/module/scanner.go
+++ b/integration_tests/.template/module/scanner.go
@@ -47,7 +47,7 @@ func RegisterModule() {
 }
 
 // NewFlags returns a default Flags object.
-func (module *Module) NewFlags() interface{} {
+func (module *Module) NewFlags() any {
 	return new(Flags)
 }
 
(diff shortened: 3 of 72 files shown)


zakird (approved):

thecsw:
Crazy because this was up for a while unreviewed -> https://github.com/zmap/zgrab2/pull/477

Issue #486, opened 2025-02-03
can not use 'local-addr' option to specify Local source address for outgoing connections


Baoxd123:
'config.go' has the "localAddr *net.TCPAddr", but 'local-addr' option was not realized to use.

PR #487, opened 2025-02-03
Realized the option of specifying the local address of zgrab2

Baoxd123:
Realized the option of specifying the local address of zgrab2, which can be verified by tcpdump, example: 'echo "172.67.161.24, renovatepattaya.com" | ./zgrab2 --local-addr=128.192.12.101:0 tls'

## How to Test

(long:"local-addr" description:"Local source address for outgoing connections (e.g. 192.168.10.2:0, port is required even if it's 0)")

In terminal 1, run zgrab2:
'echo "172.67.161.24, renovatepattaya.com" | ./zgrab2 --local-addr=128.192.12.101:0 tls'

In terminal 2, run tcpdump to verify the local address:
sudo tcpdump -i any -nn port 443
## Notes & Caveats

In the previous version of zgrab2, 'config.go' had the 'local-addr' option but could not be used. I fixed the 'local-addr' option which will be shown by 'zgrab -h'. Users can specify different local addresses, if any.

## Issue Tracking

https://github.com/zmap/zgrab2/issues/486


phillip-stephens:
Hey @Baoxd123! Thanks for opening this issue + PR. 
I was just testing out your branch and it doesn't seem to be a complete fix to the issue.
Taking the command:
```shell
/tmp/zgrab2-baoxd on  fix-local-addr ⌚ 18:21:08
$ echo "prstephens.com" | ./zgrab2 tls --local-addr="171.67.71.210:45678"
```

An initial TCP connection is opened with the correct `local-add`, but then a subsequent one is opened and used for the actual TLS connection.

image

I'm working on adding support for this so if you want to wait for that to be done, feel free. Or if you get something working for all modules, I'll review that!

phillip-stephens (changes requested):
Mentioned in prior comment, this is an incomplete fix and currently modules are ignoring this flag in at minimum the `TLS` case.

phillip-stephens:
Closing with #519

PR #497, opened 2025-03-04
Bump all dependencies

phillip-stephens:
`go get -u` and wait a bit longer for docker containers to come up before running integration tests.


phillip-stephens:
Looks like this test is failing with Snyk, it complains that `crypto` is introducing new vulnerabilities but this updates it to the latest version. I'll look into this later.

PR #505, opened 2025-03-12
Generate server certs for MQTT test container

Seanstoppable:
This generates server certificates for integration tests, rather than rely on ones checked into the repo

My org has a pretty strict policy around certificates being checked in to a repo, and discovered this when trying to merge recent changes. To work around this, I suggest we generate the certificates as part of the server startup instead.
Bonus, tests won't fail whenever the expiration of the existing certificate expires.


phillip-stephens (approved):
LGTM, thanks for this contribution!

PR #515, opened 2025-03-25
Fix SHA256 typo

hegga:
Fixed a simple SHA256 typo in the banner module, just added a missing "6"

## Notes & Caveats

Motivation was a need for correctly normalised data


phillip-stephens (approved):
Thanks for the contribution @hegga

PR #517, opened 2025-04-02
Use context.Deadline in addition to specified timeouts for connections

phillip-stephens:
Handles a number of issues/inconsistencies around how we handle timeouts/contexts when dialing connections. My goal here is to have all codepaths use our `zgrab2.Dialer` and handle all connection creation there. This should both minimize the area for bugs to creep in (forgetting to set a `ReadLimit` somewhere) and also make it easier to plug in the source IP/port work.

- Before `DialTimeoutConnectionEx()` was taking in a `context.Context`, but then not using it for anything
- Our `zgrab2.Dialer` wraps a `net.Dialer`. `net.Dialer` includes a `Timeout`, so having `zgrab2.Dialer.ConnectionTimeout` is redundant. This just sets the inner `net.Dialer.Timeout` instead of having a `ConnectionTimeout`
- Small nit - felt weird that since our dialer wraps `net.Dialer`, to have it be a named field. Then you have `zgrab2.Dialer.Dialer`. Changed the `net.Dialer` to be an anonymous structure.
- Removes the deprecated `DualStack: true`, it's now true by [default](https://pkg.go.dev/net#Dialer).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,7 @@ all: zgrab2
 # Test currently only runs on the modules folder because some of the 
 # third-party libraries in lib (e.g. http) are failing.
 test:
+	go test -v .
 	cd lib/output/test && go test -v ./...
 	cd modules && go test -v ./...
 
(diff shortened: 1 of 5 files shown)


phillip-stephens:
WIP, not ready for review.

zakird (approved):

PR #533, opened 2025-04-23
Give user IP that domain resolves to

### LINKED ISSUE #104: Domain scanning
Currently if a domain is passed in, it can be looked up again for each scan.

Would it make more sense to look it up once at the start, pass it to each call to `Scan()`, and return it in the result (so that it is possible to determine which host was actually scanned)?

phillip-stephens:
This PR does a couple things:

- ~~Changes the default DNS resolvers to a combination of Cloudflare/Google~~ Uses OS defaults
- If a user gives `--local-addr`, use that to detect IPv4/IPv6 support
    - If they don't, attempt to connect to both IPv4 and v6 resolvers to self-detect. From then on, use this info to decide to use A or AAAA records during resolution
- If the user only provides the domain, we'll fill out the IP we resolve the domain to. Additionally, we'll do this outside of the modules, so if the `MULTIPLE` module is used, the target is only resolved once.
- With `http`, we have a wrinkle of complications with re-directs. I've modified the `net.Dialer` to use a custom resolver respecting our nameservers. Additionally, I've included a mapping of `resolved_names_to_ips` in the `http` response:
```
..."redirects_to_resolved_ips":[{"redirect_name":"facebook.com","ip":"157.240.22.35:443"},{"redirect_name":"www.facebook.com","ip":"157.240.22.35:443"}]},"...
```

## How to Test

To test, I spot-checked the output on sample domains and on 10k, spot checking a few throughout the output.

## Notes & Caveats

~~It's important to note that this PR _ignores the OS-specified (ie: `/etc/resolv.conf`) file and uses Google/Cloudflare_. The user can ofc override this. Anecdotally, I saw better results using these and I think it's a reasonable choice for most people.~~ Uses OS defaults now

## Issue Tracking

Resolves #104

### DIFF
--- a/config.go
+++ b/config.go
@@ -13,20 +13,6 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-var defaultIPv4DNSResolvers = []string{
-	"1.1.1.1:53",
-	"1.0.0.1:53",
-	"8.8.8.8:53",
-	"8.8.4.4:53",
-}
-
-var defaultIPv6DNSResolvers = []string{
-	"2606:4700:4700::1111:53",
-	"2606:4700:4700::1001:53",
-	"2001:4860:4860::8888:53",
-	"2001:4860:4860::8844:53",
-}
-
 // Config is the high level framework options that will be parsed
 // from the command line
 type Config struct {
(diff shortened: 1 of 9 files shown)


zakird:
I know this is still in draft mode, but I think that we should reconsider this default
> Changes the default DNS resolvers to a combination of Cloudflare/Google

and instead use the machine's default resolver unless it's specified otherwise

zakird (approved):

PR #556, opened 2025-06-02
Bugfix - allow user to set the --dns option

phillip-stephens:
Realized the `--dns` option was basically a no-op and we weren't setting things appropriately. 

## Root Cause
The issue came to be since in `SetDefault` I wanted to avoid over-writing any pre-existing changes. So I only setup the `--dns` option on the resolver if the `Dialer` was nil. The issue is the caller initialized the dialer before calling `SetDefaults`, so effectively we never set `--dns`. 

## How to Test
A bit difficult to test, but had a DigitalOcean droplet where it seems the local DNS resolver would rate-limit you and stop resolving around 300-400 queries. Setting `--dns=1.1.1.1` had no effect, which was really surprising. With this fix, we can resolve.


zakird:
I think that there's probably a better name for this field like `--dns-resolvers=1.1.1.1`.

PR #557, opened 2025-06-02
Remove IP capability detection, enable IPv4 by default

### LINKED ISSUE #554: Do not connect to non-target IP addresses to identify IPv4/IPv6 capabilities
## Background
Users can pass in targets in two ways, by domain name or IP address. If they provide an IP address we'll need to resolve this into IPv4 and/or IPv6 addresses. The issue is which to use? How can we detect the IPv4 and IPv6 capabilities of the host? There are some hosts which have an IPv6 network address but upstream routers don't have support and therefore they can't scan IPv6 addresses. We'll call these hosts as having _local but not internet IPv6 access_.

What is done at present is at `zgrab` startup we'll attempt to connect to a static IP address `one.one.one.one` with both IPv4 and IPv6 and detect host capability in this way.

## Issues with Current Approach

This means we:

- Require internet access (`zgrab` can't be used offline)
- `one.one.one.one` can't be blocked/subject to censorship
- It's not obvious to users that scanning a given target IP address would also leak their public IP to `one.one.one.one`

## Approach
Let's just default to `--4=true` and `--6=false` and IPv6 can be opt-in. That way it's a user-driven choice.

Relates to #552

phillip-stephens:
Removes the host auto-detection code for scanning host's IPv4/v6 capabilities. Rationale is that it isn't clear to user's that ZGrab would reach out to `1.1.1.1` when they're not scanning that just to figure out IP capabilities. Additionally, it means ZGrab must have internet access to function.

## Changes
- `--use-ipv4` is functionally set to true by default
- `--use-ipv6` is set to false by default

Resolves #554


Seanstoppable:
Is this strictly for resolving hostnames?
Rather than `use-ipv4/6`, would `resolve-ipv4/6` be better terminology?

zakird (approved):

PR #572, opened 2025-06-25
Phillip/342 rate limit

### LINKED ISSUE #342: Rate limiting
There was another ticket on the topic #221 but since I'm unable to reopen I had to create a new one. Simply said, by running zgrab2 with a large list of domains, seems to result in overloading the DNS server, while the server under which I'm running zgrab2 is having ulimit set to unlimited.

{"domain":"asnpart.net","data":{"http":{"status":"connection-timeout","protocol":"http","result":{},"timestamp":"2022-01-29T22:44:12+01:00","error":"dial tcp: lookup asnpart.net on 185.12.64.1:53: dial udp 185.12.64.1:53: socket: too many open files"}}}

phillip-stephens:
Adds a per-IP and a overall DNS rate limiter to ZGrab

## Details
Both rate limits are user-configurable
- `--dns-rate-limit` - defaults to 10k requests per second to lookup names
- `--server-rate-limit` - A per-IP address rate limit, defaults to 20 connections per second per IP

Both are handled at the framework layer, so as long as a module uses our `TimeoutConnectionDialer`, and all modules do by default if they're using our default dialers, we can rate limit them.

### DNS Rate Limiter
This is a single global rate limiter for DNS lookups. Looking at how we can plug into the DNS resolution, I couldn't see a clean way that would let us use the system's (not Go's internal resolver) resolver _and_ get what resolver was being used for the lookups. Therefore, we're using a single rate limit across all. I think this is acceptable for users, they can always increase this limit if they know they're using many recursive resolvers and the load will be spread across them.

### Per-IP Rate Limit
This uses an LRU cache allowing for efficient, thread-safe lookups as well as expiring limiters that haven't been used in a while. A simple map of IP -> limiter would grow unbounded, so we need an efficient way to expire limiters that haven't been used in a while. I chose to set the bounded size of the cache to 10M (unrealistically large) but with a `10s` TTL. I've found even with an incredibly beefy VM, I didn't see cache occupancy over 10-20k. I also verified with heap profiling that we didn't have a memory bloat problem because of this, the cache fits in < 5 Mb of memory.

## Performance
I didn't notice a noticeable performance difference with this when testing with many threads and comparing to `master`.

```shell
tail -n 100000 /tmp/1m-top-domains.csv| ./zgrab2 http --max-redirects=5 --output-file=/dev/null  -t 15s --senders=5000 --dns-resolvers=1.0.0.1,1.1.1.1 
```

## Accuracy

I tested both limiters as follows:
### DNS
```shell
tail -n 100000 /tmp/1m-top-domains.csv  | ./zgrab2 http --max-redirects=5 --output-file=/dev/null -t 15s --senders=50 --dns-resolvers=1.0.0.1,1.1.1.1 --server-rate-limit=20 --dns-rate-limit=1     
```
I then used the `dns` display filter in Wireshark to ensure we were only sending 1 query a second. 

### Per-IP
```shell
yes "prstephens.com" | ./zgrab2 http --max-redirects=5 --output-file=/dev/null -t 15s --senders=50 --dns-resolvers=1.0.0.1,1.1.1.1 --server-rate-limit=1
```
I then used the Wireshark display filter `tcp.flags.syn == 1 && tcp.flags.ack == 0` to ensure we were only initiating a single connection to each IP per-second.

## Notes & Caveats

Rate-limiting by IP is the easiest to accomplish. However for avoiding bot detection, it's presently unclear whether IP is the correct Key or something like CIDR block or Autonomous System number of server IP. Leaving this for future work.

## Issue Tracking

Resolves #342


zakird (approved):

zakird:
Overall this looks good. Performance measurement is a bit hand-wavy ;) but trusting it looks good.

PR #573, opened 2025-06-25
Retry Postgres scans without TLS if the TLS connection attempt errors

UnknownEclipse:
Some Postgres protocol implementations/configurations (such as PgBouncer and AWS Redshift) return an `ErrorResponse` upon first connecting (likely due to the permissions they are configured with). This causes the scan to be marked as failed even though there is an active server running. This PR retries without SSL if that error response is received.

## How to Test

Find a host that is incorrectly failing scans, and run

```sh
./zgrab2 postgres
```

## Notes & Caveats

A note on how this failure case is handled: this change simply retries. However, it may be more optimal to exit out early with a success code but without attempting to elicit an unsupported version response. In every case I've seen the unsupported version error is never received after receiving an error from the SSL Request so this would simply skip those extra connections. I'm working on finding any examples of it being beneficial to continue with a retry.

A quick note on responses: The existing scanner does not do any kind of validation before setting `SupportedVersions`. Even the current version may sometimes include a totally unrelated error message in rare cases, but this PR will increase the amount of cases that will happen. Do we want to add a `if contains(response, "unsupported")` check to prevent that from occurring?

### DIFF
--- a/modules/postgres/scanner.go
+++ b/modules/postgres/scanner.go
@@ -358,7 +358,7 @@ func (s *Scanner) DoSSL(ctx context.Context, sql *Connection, dialGroup *zgrab2.
 }
 
 // newConnection opens up a new connection to the ScanTarget, and if necessary, attempts to update the connection to SSL
-func (s *Scanner) newConnection(ctx context.Context, t *zgrab2.ScanTarget, mgr *connectionManager, nossl bool, dialGroup *zgrab2.DialerGroup) (*Connection, *zgrab2.ScanError) {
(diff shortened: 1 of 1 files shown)


phillip-stephens:
Hey @UnknownEclipse, thanks for this! This seems very reasonable, as I was reviewing I didn't like how we had a boolean called `noSSL` and think having booleans follow `useX` makes them more readable. I made that fix since that wasn't caused by you. Let me do a quick sanity check on a Postgres scan but otherwise thanks for the contribution!

phillip-stephens:
I did a small ZGrab scan comparing this to `master` with 4k IPs that respond to `zmap` on port 5432. I found one AWS Redshift host responding like you mentioned in your Caveats. However, we at least get more info than `master` which just returns an `application-error`. So I think this is strictly an improvement, LGTM.

phillip-stephens (approved):

PR #578, opened 2025-07-18
Fix "Application Options" mutliple module parsing

### LINKED ISSUE #337: Documentation does not match the actual config format - 'Application Options' does not get parsed at all
Basically the same as #236, so it's a duplicate... But since @mzpqnxow closed the issue with this still being quite a major issue, I am reposting this.

The README.md says you can define a config file like this:
```ini
[Application Options]
output-file="output.txt"
input-file="input.txt"
[http]
name="http80"
port=80
endpoint="/"
[http]
name="http8080"
port=8080
endpoint="/"
[ssh]
port=22
```
And pass it as: `./zgrab2 multiple -c multiple.ini`.

It also says: 
>`Application Options` must be the initial section name.

Now, this is in fact a complete lie. zgrab2 does not parse this section AT ALL. You can put any other global setting in there, it is getting ignored, and there is no section in the code actually parsing this section.

I am not sure how this even happened... But as a workaround, everything you want to pass in the config file as `Application Options`, you need to pass to your config.

So, if you want to pass input and output, you need to do it like this: `./zgrab2 -f input.csv -o output.csv multiple -c multiple.ini`.
Same goes for other global options. For example, if you want to increase the size limit per scan, you need to do it as followed:
 `./zgrab2 -f input.csv -o output.csv multiple -c --read-limit-per-host 1048576 multiple.ini`.

Quite a major issue, in my eyes...

phillip-stephens:
As reported in #337, we were parsing the module-specific flags (e.g.: `--max-redirects` for the `http` module) but not the tool-wide "Application Options" in `.ini` files. Users instead **had** to resort to passing those in as CLI args. This was in contradiction to the instructions in the README.

## Prior Behavior
```ini
[Application Options] // This section is ignored
senders=2
input-file="/multiple/input.csv"

[ntp]
trigger="ntp"

[http]
trigger="http80"
```

Users must provide all Application Options as CLI, `argv` flags, in order for them to take effect. This is the workaround discussed in #337.
```shell
./zgrab2 multiple --input-file="/multiple/input.csv" --config-file="config.ini" --senders=2
```

## New Behavior
We now have a hierarchy of inputs, as suggested by @Seanstoppable on how most CLI apps work.

1. The config is populated with sane defaults (e.g. a rate limit per server IP of 20 reqs/sec) which are communicated with users through `--help` text.
2. If using the `multiple` module, the `Application Options` section defines application-wide options. If present, these override the defaults above.
3. After parsing the CLI flag options, if provided, this will override both of the above.

## How to Test

Modified the existing integration test to provide the `--input-file` only in `.ini`. This way, if the Application Options section of the INI file isn't parsed, the test won't read the input domains to scan.

Also added an ini file with a non-existent input file, and provided the test with a good input file over CLI args. In this way, we test that the CLI arg is taking precedence over the INI file.

## Notes & Caveats

~~The workaround will no longer work. Any CLI flags a user specifies as arguments (e.g. `--senders` in `./zgrab2 multiple -c "config.ini" --senders=3`) will be disregarded as the values will now come only from the `ini` file. This matches the behavior previously specified in the README but anyone depending on the workaround will need to switch.~~

The "workaround" still works since CLI flags take precedence over INI files.

## Issue Tracking

Resolves #337

### DIFF
--- a/integration_tests/multiple/bad-input-file.ini
+++ b/integration_tests/multiple/bad-input-file.ini
@@ -7,3 +7,4 @@ trigger="ntp"
 
 [http]
 trigger="http80"
+max-redirects=3
--- a/integration_tests/multiple/input.csv
+++ b/integration_tests/multiple/input.csv
@@ -1,2 +1,2 @@
-, example.com, http80
+, github.com, http80
 , time-a-g.nist.gov , ntp
--- a/integration_tests/multiple/multiple.ini
+++ b/integration_tests/multiple/multiple.ini
@@ -7,3 +7,4 @@ trigger="ntp"
 
 [http]
 trigger="http80"
+max-redirects=3
--- a/integration_tests/multiple/test.py
+++ b/integration_tests/multiple/test.py
@@ -41,12 +41,14 @@ multiple_test_root = os.path.join(zgrab_root, "integration_tests", "multiple")
 
 def validate_http(output):
     json_data = json.loads(output)
-    assert json_data["domain"] == "example.com"
+    print(json_data)
+    assert json_data["domain"] == "github.com"
     assert json_data["data"]["http"]["status"] == "success"
 
 
 def validate_ntp(output):
     json_data = json.loads(output)
+    print(json_data)
     assert json_data["domain"] == "time-a-g.nist.gov"
     assert json_data["data"]["ntp"]["status"] == "success"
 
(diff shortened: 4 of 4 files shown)


Seanstoppable:
I didn't realize I was using a workaround and do not relish updating configs. :)

A lot of tools with config files have a hierarchy they adhere to like env vars > cli flags > config file and wondering if we should do so here, even if temporarily, for users to more gradually migrate.

I.e support both now and then remove what you view as a workaround next year.

phillip-stephens:
@Seanstoppable fair point, let me play with getting that kind of support added to this next week!

phillip-stephens:
Made this change and tested locally as well as added an integration test. Once #585  is merged into `master`, tests should pass here and we can merge it.

zakird (approved):

PR #579, opened 2025-07-21
feat(http): add `SkipHost` request option

UnknownEclipse:
Adds a `SkipHost` option in `http.Request`. This skips encoding the host header, which is against the standard, but can be useful for some servers that immediately close the connection (without issuing any response) if the host header is not recognized.

## How to Test

Find a known HTTP server that returns an EOF error during scanning with no data received. Then issue an HTTP request with `SkipHost: true` and confirm that a 400 is returned.

## Notes & Caveats

This is a messy way of handling things, but some information is better than nothing at all unfortunately.

### DIFF
--- a/lib/http/request.go
+++ b/lib/http/request.go
@@ -230,6 +230,12 @@ type Request struct {
 	// domain name.
 	Host string `json:"host,omitempty"`
 
+	// SkipHost is a flag that indicates whether to skip the host header.
+	// This is against the standard, but can be useful when working with servers
+	// that immediately close the connection upon receiving a host they do not
+	// recognize
+	SkipHost bool `json:"skip_host,omitempty"`
+
 	// Form contains the parsed form data, including both the URL
 	// field's query parameters and the POST or PUT form data.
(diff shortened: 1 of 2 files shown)


phillip-stephens:
I'm not initially opposed to adding this flag, seems like a small enough change for the use case. Two questions:

1. Looks like this PR is to merge this into `feature/TLS1.3`, I'd imagine we'd this in the `main` branch?
2. I don't see anyway a CLI user can invoke this, we'll need a CLI flag to toggle this no?

UnknownEclipse:
> I'm not initially opposed to adding this flag, seems like a small enough change for the use case. Two questions:
> 
> 1. Looks like this PR is to merge this into `feature/TLS1.3`, I'd imagine we'd this in the `master` branch?
> 2. I don't see anyway a CLI user can invoke this, we'll need a CLI flag to toggle this no?

1. The initial motivation was a project that is still on `feature/TLS1.3`. I'd be happy to create a separate PR to make the changes to master though!
2. Will add that

codyprime (approved):

PR #581, opened 2025-07-24
Add support for HTTP/2 and HTTP/2 over clear-text

### LINKED ISSUE #22: Implement HTTP/2
Per @dadrian, 
>We have the code for HTTP/2, but need to resolve some differences with zcrypto/tls and crypto/tls in order to it to compile and pass tests.

phillip-stephens:
- Updates the `lib/http` package to be current with Go 1.23, cherry-picking the `zgrab2` modifications in
- Adds the `golang.org/x/net/http2` package so we can access the internals of the http2 TLS negotiation and have it use our crypto
- Following common browser convention, the default is to send an `http/1.1` request but when initiating a TLS connection, advertising HTTP/2 support in the TLS `next-protos` field. Then, if the server supports, it'll choose `h2` and the connection will be upgraded.
    - Users can use `--http2-prior-knowledge` to force HTTP/2 without TLS (also called `h2c`). This skips the `http/1.1` initial msg + upgrade to begin with HTTP/2.
    - Users can specify only `HTTP/2` or `HTTP/1.1` by setting the TLS `--next-protos`. For example, `--next-protos="h2,http/1.1" is the default, but users can remove one or the other to only use one or the other.
- Tests for servers only supporting HTTP/1.1, HTTP/2, and h2c added
    - To accomplish this, split our existing http tests into `http_smoke_test` and `http_version_tests`
    
## Testing To Be Done Before Merge
Once we iron out the CLI syntax, I'll run these tests before merging:

- [ ] http/1.1 over http
- [ ] http/2 over http
- [ ] http/1.1 over https
- [ ] http/2 over https
- [ ] http/2 OR http/1.1 with a server that only supports h2
- [ ] http/2 OR http/1.1 with a server that only supports h/1.1
- [ ] NextProtos = h2, h1.1 - Redirect from Server that only supports HTTP/1.1 to a server that supports HTTP/2, we should upgrade
- [x] --fail-http-to-https
- [x] --retry-https

## Limitations/Breaking Changes
The underlying `http` library from `net/http` modified how the HTTP protocol is represented by unpacking the `protocol` struct. See below
```json
-          "protocol": {
-            "major": 1,
-            "minor": 1,
-            "name": "HTTP/1.1"
-          },
+          "protocol": "HTTP/2.0",
+          "protocol_major": 2,
```
This means any downstream processing relying on this structure will need to be updated.

### Regression Testing
First I wanted to be sure that there were no regressions in functionality if the user forces http/1.1. Taking Cloudflare.com as an example, I ran the following on both `master` and this branch:
```shell
echo "cloudflare.com" | ./zgrab2 http --max-redirects=5 --next-protos="http/1.1" > /tmp/22-http-zgrab-testing/http-1.1/cloudflare.22
```

The only differences in the output between the two branches beyond diffs in time, `cf_ray`, and ip were the `protocol`:
```json
-          "protocol": {
-            "major": 1,
-            "minor": 1,
-            "name": "HTTP/1.1"
-          },
+          "protocol": "HTTP/1.1",
+          "protocol_major": 1,
+          "protocol_minor": 1,
```

### HTTP/2 Support/Default Behavior
As previously mentioned, the default behavior of the new `http` module is to support HTTP/2 upgrade of connections.
Comparing a default run of zgrab with `master` and current branch:
```shell
$ echo "apnews.com" | ./zgrab2 http --max-redirects=5 > /tmp/22-http-zgrab-testing/default-behvaior/apnews.master
```
```json
@@ -52,11 +52,9 @@
                 "Accept-Encoding"
               ]
             },
-            "protocol": {
-              "major": 1,
-              "minor": 1,
-              "name": "HTTP/1.1"
-            },
+            "protocol": "HTTP/1.1",
+            "protocol_major": 1,
+            "protocol_minor": 1,
             "request": {
               "headers": {
                 "accept": [
@@ -68,6 +66,9 @@
               },
               "host": "apnews.com",
               "method": "GET",
+              "protocol": "HTTP/1.1",
+              "protocol_major": 1,
+              "protocol_minor": 1,
               "url": {
                 "host": "apnews.com",
                 "path": "/",
             "age": [
-              "73"
+              "30"
             ],
@@ -159,11 +157,8 @@
               "Brightspot"
             ]
           },
-          "protocol": {
-            "major": 1,
-            "minor": 1,
-            "name": "HTTP/1.1"
-          },
+          "protocol": "HTTP/2.0",
+          "protocol_major": 2,
@@ -456,16 +451,13 @@
             }
           },
           "status_code": 200,
-          "status_line": "200 OK",
-          "transfer_encoding": [
-            "chunked"
-          ]
+          "status_line": "200 OK"
```
Other than the aforementioned protocol, the primary differences here are the age from 73 -> 30, the removal of the `keep-alive`, and addition of the `transfer_encoding` being `chunked`.
- age - refers to how long this resource was cached in a proxy server, so is expected to change.
- `keep-alive` - baked into `http/2` by default, so isn't applicable
- `transfer_encoding` - "Chunked transfer encoding is not supported in [HTTP/2](https://en.wikipedia.org/wiki/HTTP/2)" - [Source](https://en.wikipedia.org/wiki/Chunked_transfer_encoding)

### Large Scale Scanning
On a scan of 10k domains, `master` branch was within 2-3% accuracy of this branch. It seems that the differences in accuracy can be chocked up more to bot-detection and blocking than differences in the performance, since each branch was trading places in accuracy depending on in what order I ran the different branch tests.

Tested with:
```shell
head -n 10 ~/zdns/10k.domains| ./zgrab2 http --max-redirects=5  | grep -c "200 OK"
```

Performance Was also comparable, all runs were within 5s of each other at around 75s for 10k domains.

## Issue Tracking
Resolves #22


zakird (approved):

zgrab2_schemas/zgrab2/http.py
```diff
@@ -151,13 +154,9 @@
         "status_line": String(),
         "status_code": Unsigned32BitInteger(),
         # lib/http/protocol.go: http.Protocol
-        "protocol": SubRecord(
```
Seanstoppable, line 154: So, I do kind of get this, but at this point the change is a non-backwards compatible change for parsing, and wondering if it is really worth it.
For example, I know that I will need to:
Upgrade my parsing code to add these fields so it doesn't blow up (which is my own choice so I know I am missing handling for fields).
Write some code for the parser to check both the old nested structure to see if it exists, and if not use the new structure, so I can migrate.
Then I can pull these changes into zgrab2 and start scans with this change.

If I don't follow this chain of steps, it means that at best I blow up my pipeline and at worst I drop information.

modules/http/scanner.go
```diff
@@ -373,55 +382,82 @@ func getHTTPURL(https bool, host string, port uint16, endpoint string) string {
 
 // NewHTTPScan gets a new Scan instance for the given target
 func (scanner *Scanner) newHTTPScan(ctx context.Context, target *zgrab2.ScanTarget, useHTTPS bool, dialerGroup *zgrab2.DialerGroup) *scan {
+	cfg := &tls.Config{
+		InsecureSkipVerify: true,
+		ServerName:         "localhost",
```
Seanstoppable, line 387: I am guessing that a ServerName MUST be supplied and we are doing this for IF we are upgraded to h2, and to ignore validation because that is what we want?

modules/http/scanner.go
```diff
@@ -373,55 +382,82 @@ func getHTTPURL(https bool, host string, port uint16, endpoint string) string {
...
+	ret.client.UserAgent = scanner.config.UserAgent
 	if scanner.config.TargetTimeout != 0 {
 		ret.globalDeadline = time.Now().Add(scanner.config.TargetTimeout)
 	}
-	ret.transport.DialTLS = func(network, addr string) (net.Conn, error) {
-		deadlineCtx, cancelFunc := ret.withDeadlineContext(ctx)
-		conn, err := dialerGroup.GetTLSDialer(deadlineCtx, target)(network, addr)
-		if err != nil {
-			return nil, fmt.Errorf("unable to dial target (%s) with TLS Dialer: %w", target.String(), err)
+	if scanner.config.HTTP2PriorKnowledge {
```
Seanstoppable, line 403: This is semi-confusing because for http we'd set:
scanner.config.NextProtos = "http/1.1"
and to force h2 over https we'd set
scanner.config.NextProtos = "h2"
But for h2 without TLS we'd set another option entirely

I realize part of the problem is NextProtos is part of the tls config, but wondering if there is a more consistent way we can do this.

Seanstoppable:
This adds a regression in behavior:

echo 69.165.213.154 | ./zgrab2 http --endpoint='/'  -p 80

returns with an "unknown-error": "error populating response body after 0 bytes: http: read on closed response body"
Where before it would be a success with appropriate body

zgrab2_schemas/zgrab2/http.py
phillip-stephens: This is true, though I'm hesitant to stray from the net/http libraries implementation.

I'd argue that a) we're still on `v0` and so breaking changes are acceptable and b) diverging from net/http by keeping the old protocol struct format means any code that was updated in the new http implementation will need to be modified to work with the old protocol struct layout. 

It's unfortunate but IMO worth it to gain this new feature set. Unless you see another way

Seanstoppable: That is fair as well. The fewer diverging changes from upstream the easier the maintenance. Dealing with my own downstream changes for HTTP is its own headache.

I'd push back a bit gently on the 'version 0' thing a bit, that while strictly true that some recent comments about the dialergroup indicates that some downstreams are not expecting major breaking changes.

In this example the changes are kind of trivial, and I will certainly buy the maintenance ease over dying on the backwards compatibility hill.

Seanstoppable:
Yeah, none of the integration tests check for scan status.
Here is a little PR that adds a bit, even if it doesn't catch this particular regression
https://github.com/zmap/zgrab2/pull/606

modules/http/scanner.go
phillip-stephens: Yeah I agree, thinking a bit more about it I'm thinking we can just follow curl's syntax

We add 2 flags: `--use-http2=true` and `--use-http1.1=true` and remove `--http2-prior-knowledge`

These will control both the ALPN field in TLS handshakes (both on following redirects to `https://` sites and on initial connection when `--use-https=true`) and what HTTP version is used for the initial connection or whenever we don't have a TLS connection.

There's the settings we use on the initial connection and then settings we use when following re-directs. For example, when using `zgrab http --use-https=false --max-redirects=5`, the initial request will be made over plain-text HTTP but on re-directs we may need to follow a `https://` URL and we'll need TLS.NextProto set correctly.

I'm thinking these as defaults. If the user directly specifies `--next-protos=X` that will take precedence over any ALPN settings listed below.

- Default case - Initial plain-text HTTP/1.1 w/ both HTTP/2 and HTTP/1.1 advertised in ALPN on re-directs
	- `zgrab --use-https=false --use-http2=true --use-http1.1=true`
	- ALPN set on re-directs to `h2,http/1.1`
	- Initial HTTP/1.1 request (follows browser conventions)
- HTTP/2 Only
	- `zgrab --use-https=true --use-http2=true --use-http1.1=false`
	- ALPN set to `h2`
	- Initial TLS connection with HTTP/2 request
- HTTP/2 over Cleartext (`h2c`)
	- Equivalent to curl's `--http2-prior-knowledge`
	- `zgrab --use-https=false --use-http2=true --use-http1.1=false`
	- ALPN only on re-directs, set to `h2`
	- Initial connection is a plain-text HTTP2 request
- HTTP/1.1 Only over TLS
	- `zgrab --use-https=false --use-http2=false --use-http1.1=true`
	- Initial connection is a plain-text, HTTP/1.1 request
	- ALPN only on re-directs, set to `http/1.1`

Issue #583, opened 2025-07-29
Siemens S7 module fails with io-timeout on remote devices


zsbahtiar:
Hi, I'm getting consistent timeout errors when trying to scan Siemens S7 devices over the internet. Local scans work fine but remote ones always fail.
```
echo xxx.xxx.xxx.xxx | zgrab2 siemens
Returns:
  {
    "status": "io-timeout",
    "error": "could not send s7 request and read response: read tcp 192.168.0.8:50148->xxx.xxx.xxx.xxx:102: i/o timeout"
  }
```

- Confirmed port 102 is open: nc -zv xxx.xxx.xxx.xxx 102 works fine
- Tried increasing timeouts: --connect-timeout=30s --target-timeout=120s but no difference
- The device is definitely a Siemens PLC (works with other S7 tools)

Looking at the code, sendRequestReadResponse() in modules/siemens/s7.go uses ReadAvailable() which has a hardcoded 10ms timeout.
This is way too short for internet scanning - my device takes ~500ms to respond.

 The fix should be switching to ReadAvailableWithOptions() with a more reasonable timeout like other modules use

PR #584, opened 2025-07-29
feat(siemens): add custom readtimeout by flag

zsbahtiar:
The siemens module was failing with io-timeout errors because it uses ReadAvailable() which has a hardcoded 10ms timeout - way too short for real network conditions. Changed it to use ReadAvailableWithOptions() we can custom the timeout instead

## How to Test

1.  Find a Siemens S7 device accessible over the internet (port 102)
2. Run: echo <target-ip> | zgrab2 siemens
3. Should now return device info instead of timing out

 Tested on a real SIMATIC 300 device and it works now.

## Notes & Caveats

Hit this issue while scanning S7 devices over the internet. The 10ms timeout is fine for local networks but fails on anything with decent latency. Chose custom timeout to match what other modules are doing (memcached uses 500ms-1s, etc).
Only add the flag read timeout for S7 protocol responses, doesn't affect the overall connection timeout.

## Issue Tracking

https://github.com/zmap/zgrab2/issues/583


modules/siemens/scanner.go
```diff
@@ -18,6 +19,7 @@ import (
 type Flags struct {
 	zgrab2.BaseFlags `group:"Basic Options"` // TODO: configurable TSAP source / destination, etc
 	Verbose          bool                    `long:"verbose" description:"More verbose logging, include debug fields in the scan results"`
+	ReadTimeout      time.Duration           `long:"read-timeout" default:"10ms" description:"Timeout for reading S7 responses"`
```
phillip-stephens, line 22:
```suggestion
	ReadTimeout      time.Duration           `long:"read-timeout" default:"500ms" description:"Timeout for reading S7 responses"`
```
Our pattern is usually to give the user sane defaults with possible customization. I think a default value of 500 ms feels better for internet-wide scanning.

phillip-stephens (commented):
Seems very reasonable, I should go inventory what other modules are using `ReadAvailable` as well, might want to make that private since it feels like a bit of a foot-gun.

If you agree with this small tweak to the default, this LGTM.

phillip-stephens (approved):

zsbahtiar:
Thanks @phillip-stephens !

zsbahtiar:
Thanks @phillip-stephens !

phillip-stephens:
Thanks for the fix!

PR #596, opened 2025-08-29
Add port to ScanResponse output

### LINKED ISSUE #590: Port not printed as part of export
In my infrastructure, I append a metadata line to the end of a scan file that has a few fields, but MOSTLY to capture the port I am scanning when processing data.

As I am doing more with the `muliple` module, this concept makes less sense, since in theory/in the example of multiple, a scan might cover multiple ports.

However, port is not part of the output, so the only way to really 'track' which port goes with which scan result appears to be to keep track of the scan name.

We also have Port on the Grab object, but it is never captured and, per above, we probably don't want to capture it there vs capturing it on the ScanResult object.

I see that ScanTarget has a Port, but I don't see how it ever really gets set. At least with the CLI I don't see how I would specify a port on the target level, rather than port on the config level.

Seanstoppable:
Since the `multiple` module lets us scan multiple ports on a single ip, add port to the ScanResponse so we can differentiate

Resolves #590

### DIFF
--- a/scanner.go
+++ b/scanner.go
@@ -97,5 +97,5 @@ func RunScanner(ctx context.Context, scanner Scanner, mon *Monitor, target ScanT
 		err = &errString
 	}
 	resp := ScanResponse{Result: res, Port: target.Port, Protocol: scanner.Protocol(), Error: err, Timestamp: t.Format(time.RFC3339), Status: status}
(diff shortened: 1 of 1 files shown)


phillip-stephens:
Thanks for this, I agree this is a valuable UX improvement.

PR #598, opened 2025-09-05
Remove unused Help function

Seanstoppable:
Empty defaults should probably be returned by the helper text of the option
This isn't used in practice, so just clear up all the empty bits

### DIFF
--- a/module.go
+++ b/module.go
@@ -226,9 +226,6 @@ type ScanModule interface {
 // ScanFlags is an interface which must be implemented by all types sent to
 // the flag parser
 type ScanFlags interface {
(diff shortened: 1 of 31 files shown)


phillip-stephens:
So I did some digging separately, and found that the CLI flags library we use didn't have implementation for `Help()`. That being said, I think `Help()` is the perfect place to put module-specific info. I fixed our fork of `flags` to support `Help()` and added text for `http` [here](https://github.com/zmap/zgrab2/blob/master/modules/http/scanner.go#L160). (`Help()` is called when a module is invoked with `--help` liike `./zgrab2 http --help`)

With that said, I'm going to close this.

PR #599, opened 2025-09-05
Make verbose a base flag

Seanstoppable:
This just cleans up verbose a bit
Most modules don't implement it, but everything still has the option to without having additional boilerplate in the individual module args


phillip-stephens (approved):
LGTM, thanks for this

PR #629, opened 2025-10-30
memcached boundry check bug fix + Checks for magic byte and opcode validations

merterdemir:
memcached boundry check bug fix + Checks for magic byte and opcode validations

- Magic Code and OpCode validation added for the binary memcached.
- When there is a response (but not a valid memcached response), ParseResponse fails to properly validate the key and value, as it parses wrong key lengths and total body lengths from an invalid response; resulting in a ```panic: runtime error: slice bounds out of range [:21608] with capacity 4864 [recovered]``` (e.g., an HTTP response) error.

## How to Test

A small 50M scan on port 80 will shortly fail and crash the ZGrab2.


phillip-stephens:
I want to a bit more testing to be sure we can find ascii only and binary only hosts.

phillip-stephens (approved):
LGTM, did some sanity tests, doesn't look like we have any regressions compared to main branch and we benefit from the added safety checks! Thanks @merterdemir for the contribution!

PR #680, opened 2026-03-03
Allow TLS scanners to downgrade to plaintext if handshake fails

Seanstoppable:
Especially for banner, when we have less of an expectation for WHAT might be listening, being able to try a TLS handshake, and downgrading to plaintext spares me from having to run a seperate scan or having to do a more complicated "multiple" module setup just to capture a large mix of errors
Other protocols added for consistency


phillip-stephens (approved):
This seems useful and looks great to me. Apologize for the delay in getting to this, appreciate the effort!

I'll fix up the CI and we can get this merged.

phillip-stephens:
As soon as #704 is merged, the CI should get fixed. Digging into that failing CI made me realize we should add a `--tls-handshake-timeout` to be shorter than the `--target-timeout` for the cases where a target doesn't respond to TLS.

PR #704, opened 2026-05-11
Add TLS Timeout Configuration Option

phillip-stephens:
Previously, the `--connect-timeout` was only being used to set the initial timeout on TCP connection. So if we wanted to scan for HTTPS on a server that responded immediately to our TCP connection but didn't respond to a TLS handshake, we'd have to wait the entire `--target-timeout` for it to time out, 60s by default.

IMO, this leads to an undesirable outcome: servers that respond to TCP but ignore TLS will wait the full `--target-timeout=60s` before timing out. This is a lot of wasted time spent waiting.

## Testing
I added generated tests for the following conditions and ensured we had the correct timeout behavior between `--connect-timeout`, `--tls-handshake-timeout`, and `--target-timeout`.
- Server doens't respond to TCP SYN
- Server responds to TCP SYN but not TLS
- Server responds to TLS handshake successfully
- Server responds to TLS handshake, reads the initial HTTP query, and then doesn't respond

### Performance
Looks like this significantly speeds up TLS scans with no meaningful drop-off in accuracy. 

Note the almost 2x increase in scan speed

```shell
$ tail -n 1 scan-results/*.status
==> scan-results/zgrab2-branch.status <==
00h:03m:05s; Scan Complete; 100000 targets scanned; 539.27 targets/sec; 71.7% success rate

==> scan-results/zgrab2-master-2.status <==
00h:05m:22s; Scan Complete; 100000 targets scanned; 309.65 targets/sec; 71.7% success rate

==> scan-results/zgrab2-master.status <==
00h:05m:25s; Scan Complete; 100000 targets scanned; 307.50 targets/sec; 71.8% success rate
```


zakird:
Hmm, I'm not sure if I agree with this offhand. It's a bit undefined whether TLS should be included, but, by default it's not in many contexts and it's useful to be able to set. In a research setting, we want this to be explicit, which says to me that we should have a different configuration option. This would also more closely match Go's semantics too I think (e.g., Go's http implementation lets you independently set DialTimeout, TLSHandshakeTimeout, and ResponseHeaderTimeout)

phillip-stephens:
@zakird That makes sense, I think adding an additional config option makes sense. It was the case before that since we only had `--connect-timeout=10s` and `--target-timeout=60s`, if a host quickly responded to a TCP handshake but then didn't respond to a TLS handshake, we'd pause for 60s. That feels incredibly long, but I take the point about wanting to be explicit. I'll add a `--tls-handshake-timeout=20s` to get some speedup and be explicit.

phillip-stephens:
@zakird ready for you when you have a moment

zakird (approved):

PR #710, opened 2026-05-13
fix: add bounds checks to MySQL packet parsing functions

Seanstoppable:
Address multiple security vulnerabilities in MySQL wire protocol parsing that could be triggered by malicious server responses:

- readNulString: return error when NUL terminator is missing instead of panicking with index -1
- readHandshakePacket: add len(rest) bounds check after readNulString before accessing fixed offsets; fix byte arithmetic overflow in part2Len calculation and comparison
- readERRPacket: add bounds checks before slicing body and rest
- readLenInt: fix 0xfd branch reading 4 bytes (body[1:5]) when only 3 are validated; use manual byte assembly instead 
- readLenString: fix off-by-one returning rest[length+1:] instead of rest[length:] — MySQL length-encoded strings have no terminator

## How to Test

## Notes & Caveats

My workplace is expanding security scans for software we use. Several modules have issues like these where we aren't doing boundary checks when parsing input

### DIFF
--- a/lib/mysql/mysql.go
+++ b/lib/mysql/mysql.go
@@ -329,10 +329,21 @@ func (p *HandshakePacket) MarshalJSON() ([]byte, error) {
 }
 
 func (c *Connection) readHandshakePacket(body []byte) (*HandshakePacket, error) {
+	if len(body) < 1 {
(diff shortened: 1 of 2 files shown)


phillip-stephens (approved):
LGTM, thanks for these improvements

PR #711, opened 2026-05-13
fix: add bounds checks to Siemens S7/COTP packet parsing

Seanstoppable:
Fix two security issues in modules/siemens/messages.go:

1. S7Packet.Unmarshal: checked only len(bytes) < 1 before reading bytes[1], bytes[4:6], bytes[6:8], bytes[8:10], and bytes[10:12]. A truncated S7 response from a malicious server causes an out-of-bounds panic, crashing the entire scan process. Fix: require len(bytes) >= 10 upfront, and >= 12 for ACK/RESPONSE PDU types before reading the error field.

2. COTPDataPacket.Unmarshal: int(headerSize+1) performed the +1 in byte arithmetic before casting to int. When headerSize=0xFF, byte(0xFF+1) wraps to 0, bypassing the bounds check entirely. Fix: cast headerSize to int first, then add 1.

## How to Test

## Notes & Caveats

My workplace is expanding security scans for software we use. Several modules have issues like these where we aren't doing boundary checks when parsing input

### DIFF
--- a/modules/siemens/messages.go
+++ b/modules/siemens/messages.go
@@ -139,9 +139,9 @@ func (cotpDataPacket *COTPDataPacket) Unmarshal(bytes []byte) error {
 		return errInvalidPacket
 	}
 
-	headerSize := bytes[0]
+	headerSize := int(bytes[0])
 
-	if int(headerSize+1) > len(bytes) {
+	if headerSize+1 > len(bytes) {
 		return errInvalidPacket
 	}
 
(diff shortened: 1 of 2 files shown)


phillip-stephens (approved):
LGTM, thanks for the fix! @Seanstoppable are the tools you're using to audit the various modules open source/something we could integrate into our CI/linters?

Seanstoppable:
@phillip-stephens unfortunately not. I am in a pilot program for some AI based security review.
BUT! Based on these, some of the problems are probably caught by fuzz testing, and I am working on a PR/seeing if it would have picked up these problems + any others.

phillip-stephens:
Oh amazing, really appreciate it!

PR #712, opened 2026-05-13
fix: return errors on BACnet tag mismatches instead of nil

Seanstoppable:
Three BACnet parsing functions (readInstanceNumber, readVendorID, readStringProperty) used a pattern where ReadByte() success was checked alongside a value comparison in a single if-statement:

    if openByte, err = buf.ReadByte(); openByte != 0x3e { return }

When ReadByte() succeeds but the value doesn't match, err remains nil and the function returns with no error — silently accepting malformed input. Callers checking only err treat the result as valid with zero-value fields.

Fix: separate the I/O error check from the value check, returning errInvalidPacket when the tag byte doesn't match the expected value.

## How to Test

## Notes & Caveats

My workplace is expanding security scans for software we use. This one is more of a consistency issue where we aren't necessarily returning the correct error.

### DIFF
--- a/modules/bacnet/objects.go
+++ b/modules/bacnet/objects.go
@@ -68,16 +68,25 @@ func readInstanceNumber(b []byte) (leftovers []byte, instanceNumber uint32, err 
 	buf := bytes.NewBuffer(b)
 	leftovers = b
 	var openByte, appByte, closeByte byte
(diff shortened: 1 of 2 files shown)


phillip-stephens (approved):
LGTM, thanks for the contribution. I'll fix up that lint.

PR #713, opened 2026-05-13
Redis Boundary checks

Seanstoppable:
Add bounds checks to prevent two denial-of-service vulnerabilities in the Redis protocol parser:

1. readRedisArray: reject negative element counts and cap at 1M elements to prevent OOM from malicious servers sending huge array sizes.

2. ReadRedisValue: add recursion depth tracking (max 32 levels) to prevent stack overflow from deeply nested array responses.

We endeavor to return as much data as possible in scenarios this happens, as well as an appropriate error.


phillip-stephens (approved):
Seems good to me, thanks!

PR #717, opened 2026-05-18
fix: initialise dnsRateLimiter to default to prevent nil panic

jwilliams-vc:
## Problem

`dnsRateLimiter` is declared as a nil `*rate.Limiter` in `scanner.go` and only assigned inside `ValidateAndHandleFrameworkConfiguration`. Library consumers that manage their own configuration and don't call that function will have a nil limiter at runtime.

When such a caller scans a target whose address resolves to a hostname — for example, when the HTTP scanner's transport follows a redirect to a `https://some-domain.com/...` URL — `lookupIPs` is called, which calls `dnsRateLimiter.Wait()` on the nil receiver:

```
panic: runtime error: invalid memory address or nil pointer dereference
...
golang.org/x/time/rate.(*Limiter).wait(0x0, ...)
    golang.org/x/time/rate/rate.go:254
github.com/zmap/zgrab2.(*Dialer).lookupIPs(...)
    conn.go:354
```

Because the panic originates inside a goroutine spawned by `(*Transport).startDialConnForLocked`, a `defer recover()` in the calling goroutine cannot catch it. The entire process crashes.

## Fix

Initialise `dnsRateLimiter` at declaration time with the same default value (10 000 req/s) that `ValidateAndHandleFrameworkConfiguration` would set. That function still overwrites the value when called, so behaviour for callers that do invoke it is unchanged.

## Alternative considered

Adding `if dnsRateLimiter != nil` guards inside `lookupIPs` would also prevent the panic but would silently skip rate limiting — the initialisation approach is safer and makes the intent clearer.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/scanner.go
+++ b/scanner.go
@@ -17,7 +17,11 @@ var orderedScanners []string
 var defaultDialerGroupToScanners map[string]*DialerGroup
 var defaultDialerGroupConfigToScanners map[string]*DialerGroupConfig
 var ipRateLimiter *ratelimit.PerObjectRateLimiter[netip.Addr]
-var dnsRateLimiter *rate.Limiter
+// dnsRateLimiter is initialised to the default rate here so that lookupIPs
+// never panics on a nil receiver when ValidateAndHandleFrameworkConfiguration
+// has not been called. ValidateAndHandleFrameworkConfiguration overwrites this
+// with the user-supplied value when invoked.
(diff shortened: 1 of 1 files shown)


phillip-stephens (approved):
Thanks for this! I went ahead and just consolidated all our initializations into this section for succinctness.

PR #727, opened 2026-06-02
feat: roll out SCAN_HANDSHAKE_ERROR to IMAP, POP3, FTP, ManageSieve, MySQL, Postgres

apowis:
## Summary

Follow-on to #718, which introduced `SCAN_HANDSHAKE_ERROR` and `SCAN_POST_TLS_APPLICATION_ERROR` for SMTP. This PR rolls the same error-status alignment out to the remaining TLS-using modules.

**Key decision from #718 review:** `SCAN_POST_TLS_APPLICATION_ERROR` is only appropriate for protocols with application-layer exchanges both _before and after_ the TLS handshake (SMTP, ManageSieve). All other modules (IMAP, POP3, FTP, MySQL, Postgres) are TCP→TLS→App and keep `SCAN_APPLICATION_ERROR` for app-level errors — only TLS handshake failures move to `SCAN_HANDSHAKE_ERROR`.

### Per-module changes

| Module | TLS handshake failure | Post-TLS app error |
|---|---|---|
| IMAP | `SCAN_HANDSHAKE_ERROR` | kept `SCAN_APPLICATION_ERROR` |
| POP3 | `SCAN_HANDSHAKE_ERROR` | kept `SCAN_APPLICATION_ERROR` |
| FTP | `SCAN_HANDSHAKE_ERROR` | N/A |
| ManageSieve | `SCAN_HANDSHAKE_ERROR` | `SCAN_POST_TLS_APPLICATION_ERROR` |
| MySQL | `SCAN_HANDSHAKE_ERROR` | kept `SCAN_APPLICATION_ERROR` |
| Postgres | `SCAN_HANDSHAKE_ERROR` | kept `SCAN_APPLICATION_ERROR` |

In all cases, any partial `TLSLog` (including `HandshakeCompletedSuccessfully`) is now captured before returning the error.

### Additional fixes

- **FTP `GetFTPSCertificates`**: return type changed from `error` to `*zgrab2.ScanError` so the caller propagates the exact status without going through `TryGetScanStatus`.
- **FTP implicit TLS (bugfix)**: `TLSLog` was being written into a local `results` variable via a `defer`, but the return value was `&ftp.results` (a value copy made before the defer ran) — so `TLSLog` was silently `nil` on every implicit-TLS success. Fixed by capturing it directly before initialising `ftp`.
- **Postgres `DoSSL`**: stores the `TLS` connection even on handshake failure so `GetTLSLog` can surface partial handshake data. Also fixes an unintended retry: the SSL-retry loop checked `connectErr.Status == SCAN_APPLICATION_ERROR` to decide whether to fall back to non-SSL — a `DoSSL` failure was incorrectly triggering that retry since it was previously `SCAN_APPLICATION_ERROR`. Now it's `SCAN_HANDSHAKE_ERROR` and the retry no longer fires.

### DIFF
--- a/modules/ftp/scanner.go
+++ b/modules/ftp/scanner.go
@@ -275,6 +275,7 @@ func (scanner *Scanner) Scan(ctx context.Context, dialGroup *zgrab2.DialerGroup,
 	if err != nil {
 		return zgrab2.TryGetScanStatus(err), nil, fmt.Errorf("error opening connection to target %v: %w", target.String(), err)
 	}
+	defer func() { zgrab2.CloseConnAndHandleError(conn) }()
 	if scanner.config.ImplicitTLS {
 		tlsWrapper := dialGroup.TLSWrapper
 		if tlsWrapper == nil {
@@ -301,7 +302,6 @@ func (scanner *Scanner) Scan(ctx context.Context, dialGroup *zgrab2.DialerGroup,
 	if tlsConn, ok := conn.(*zgrab2.TLSConnection); ok {
 		results.TLSLog = tlsConn.GetLog()
 	}
-	defer func() { zgrab2.CloseConnAndHandleError(conn) }()
 	ftp := Connection{conn: conn, config: scanner.config, results: results}
 	is200Banner, err := ftp.GetFTPBanner()
 	if err != nil {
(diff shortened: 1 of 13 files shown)


phillip-stephens (approved):
LGTM, thanks for this contribution!

PR #730, opened 2026-06-04
fix: recover from scanner panics without killing the process

Seanstoppable:
The deferred recover() in grabTarget() was re-panicking after logging, which propagates up to the worker goroutine in Process() that has no recover. A single panic in any scanner (e.g. nil deref parsing a malformed server response) would terminate the entire zgrab2 process, killing all in-flight workers and preventing buffered output from flushing.

Changes:
- Remove re-panic; record a SCAN_UNKNOWN_ERROR result instead
- Log the full stack trace via debug.Stack() for diagnostics
- Use the captured 'name' parameter instead of the loop variable 'scannerName' which may have advanced past the panicking scanner

### DIFF
--- a/processing.go
+++ b/processing.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"math/rand"
 	"net"
+	"runtime/debug"
 	"sync"
 
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


phillip-stephens (approved):
LGTM, thanks for this!

PR #732, opened 2026-06-08
Adds module for Omron Fins protocol

Seanstoppable:
Another internally maintained module that I am contributing upstream as I attempt to normalize towards maintstream zgrab2

Unfortunately, not much in the way of emulators so integration tests against a real system, but there are existing unit tests


phillip-stephens (approved):
Tested, thanks for the contribution!

PR #754, opened 2026-07-17
fix(utility): don't cap ReadAvailable's first read at the drain timeout

codyprime:
ReadAvailable wraps around ReadAvailableWithOptions, using pre-defined default values for convenience. The default readTimeout is only 10ms.

After 9572dd0e903c32baea0136d9d3ac973caa4c205b, the readTimeout is applied to the first read on the conn, not just subsequent reads. This means very short read values (default is 10ms) will fail most of the time, for non-local connections. This impacts dnp3, fox, rdp; their first read had an effective timeout of 10ms. I expect the success rate for these protocols on main is roughly 0%.

It probably also impacts modules that use ReadAvailableWithOptions, if they specify relatively short timeouts (memcached, jarm, telnet all specify 500ms)

Bound the first read by totalTimeout instead (falling back to the conn's session timeout); readTimeout now applies only to the drain reads. This has the added benefit of honoring the timeout value passed to a scanner for the first read, if the scanner passes 0 in for the total timeout.

If no total timeout is provided, rather than defaulting to 1s, it now defaults to DefaultSessionTimeout (becoming the only user of DefaultSessionTimeout I believe). This seems appropriate, given the description for it: "the default maximum time a connection may be used when no explicit value is provided"

## How to Test

Pick a DNP3 server that is not local.  (I won't provide an example one here, since this is a SCADA service). 

### Pre-fix:
```bash
echo a.b.c.d \
   | ./zgrab2 dnp3 --port=20000 --blocklist-file=/dev/null --target-timeout 10 \
   | jq -c '.data.dnp3 | {status, raw: .result.raw_response}'
```

```json
INFO[0000] started grab at 2026-07-17T17:17:51-04:00    
INFO[0000] finished grab at 2026-07-17T17:17:51-04:00   
00h:00m:00s; Scan Complete; 1 targets scanned; 3.10 targets/sec; 0.0% success rate
{"status":"io-timeout","raw":null}
{"statuses":{"dnp3":{"successes":0,"failures":1}},"start":"2026-07-17T17:17:51-04:00","end":"2026-07-17T17:17:51-04:00","duration":"322.716822ms","zgrab_cli_parameters":"./zgrab2 dnp3 --port=20000 --blocklist-file=/dev/null --target-timeout 10","num_targets_scanned":1}
```

### Post-fix:
```bash
echo a.b.c.d \
   | ./zgrab2 dnp3 --port=20000 --blocklist-file=/dev/null --target-timeout 10 \
   | jq -c '.data.dnp3 | {status, raw: .result.raw_response}'
```

```json
INFO[0000] started grab at 2026-07-17T17:17:58-04:00    
INFO[0000] finished grab at 2026-07-17T17:17:58-04:00   
00h:00m:00s; Scan Complete; 1 targets scanned; 2.81 targets/sec; 100.0% success rate
{"status":"success","raw":"BWQFCwAAAQC68A=="}
{"statuses":{"dnp3":{"successes":1,"failures":0}},"start":"2026-07-17T17:17:58-04:00","end":"2026-07-17T17:17:58-04:00","duration":"355.437127ms","zgrab_cli_parameters":"./zgrab2 dnp3 --port=20000 --blocklist-file=/dev/null --target-timeout 10","num_targets_scanned":1}
```


zakird (approved):

zakird:
@codyprime can you rebase? Apparently I don't have rights to do on your branch? Once, done, I'll merge

PR #769, opened 2026-09-09
ci: pin goimports for Go 1.25

Seanstoppable:
## Summary
Pins goimports to a release compatible with the repository's Go 1.25 CI environment.

## Changes
- Replace `goimports@latest` with `goimports@v0.49.0` in the lint workflow.
- Restore lint execution after `golang.org/x/tools@latest` raised its minimum Go version to 1.26.

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -69,7 +69,7 @@ jobs:
           go-version: ${{ steps.go_version.outputs.GO_VERSION }}
       - name: Other lint
         run: |
-          go install golang.org/x/tools/cmd/goimports@latest
+          go install golang.org/x/tools/cmd/goimports@v0.49.0
           output=$(goimports -d -local "github.com/zmap/zdns" ./)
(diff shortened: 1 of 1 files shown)


Eljees:
Drive-by on the line this touches, since it is the same one: the `-local` prefix on the next line names `github.com/zmap/zdns` while this module is `github.com/zmap/zgrab2`, so the local-import group the check means to enforce is never matched here. It came over with the copy of zdns's job.

It changes nothing today. On current `master`, with goimports v0.49.0, both spellings produce an empty diff:

```
$ goimports -d -local "github.com/zmap/zdns" ./      -> no output
$ goimports -d -local "github.com/zmap/zgrab2" ./    -> no output
```

So it is safe to correct in the same breath, and after that the job actually guards the grouping it is named for. Your pin is the right call either way; the same break is on zdns, where the lint job has been red since 9 September, and I sent the same v0.49.0 pin there as zmap/zdns#635.

zakird (approved):

PR #770, opened 2026-09-11
msmq: enterprise-scale output, anonymous-handshake security model, correlation tool

manbeastdamion:
## Summary
- Fixes a real correctness bug in the `msmq` module: the little-endian bit-order of the EstablishConnection `OperatingSystem` field was wrong (confirmed against Microsoft's own packet capture and ~30 real-world hosts), which also caused anomalous/non-compliant responders to be reported at the same confidence as genuine ones.
- Adds a graduated anonymous-access security model to `msmq` (`anonymous_handshake` / `anonymous_protocol_access` / `anonymous_resource_access`) — the latter two always `not_tested`, since this passive scanner never attempts queue/message/resource access.
- Replaces embedded prose with centralized evidence/finding codes (`MSMQ_*`) and a descriptions map, and adds hash-based raw-evidence references (`sha256`/`length`/`inline`) instead of mandatory inline blobs.
- Adds three output modes to `msmq` — `debug` (default, unchanged wire format for backward compatibility), `standard`, and `compact` (~66% smaller, <1KB/asset at scale) — via `--output-mode`.
- Adds `lib/attribution`, a small shared vocabulary (`Confidence`, `Evidence`, `Detection`, `Finding`, `OSIdentification`, `VulnerabilityAssessment`) so `msmq` and future modules don't each invent their own confidence/evidence model.
- Adds `correlate`, a new host-level correlation tool/library that combines `msmq`/`msrpc`/`smb`/`rdp` results from `zgrab2 multiple` into a weighted OS-family/vulnerability assessment — never inferring a version without real evidence (e.g. NTLM), and never treating `closed`/`timeout`/`not_scanned` as "not installed".
- Adds `cmd/zgrab2-correlate`, a thin CLI wrapper for the `correlate` package.

## Test plan
- [x] `go build ./...`, `go vet`, `gofmt -l` all clean
- [x] `modules/msmq`: 24 tests + fuzz target, including a dedicated regression test that an anomalous response can never report higher security confidence than the underlying protocol evidence
- [x] `correlate`: 12 tests covering MSMQ-only/+RPC/+RPC+SMB/+RDP-unavailable, WinRM (`not_scanned`, never spoofable), NTLM-is-the-only-version-path, no-escalation-beyond-evidence
- [x] Benchmarked at 32/1,000/100,000/1,000,000 synthetic targets: compact mode ≈934 bytes/asset (vs ≈3810 for debug), ~3.2x faster to serialize
- [x] Live-validated against ~30 real-world MSMQ hosts across multiple passes, including one consistently anomalous/non-compliant responder used as a regression fixture
- [ ] Maintainer review of the new output-mode schema and correlation-layer design


manbeastdamion:
Please review and approve

manbeastdamion:
any update guys on the review

